PR #3, opened 2020-07-13
Fix issue loading device from /tally if only one device is registered

johankj:


josephdadams:
Good catch! Thank you!

PR #5, opened 2020-08-10
Call NoSleep from within a user input event handler

albertorighetto:
noSleep.enable() "must be wrapped in a user input event handler e.g. a mouse or touch handler" in order to work.
It was not working properly on some older Android 4.1 devices because it was called inside on onLoad().
This will fix it


josephdadams:
Thank you!

PR #19, opened 2020-11-23
Added GUI selection of OBS addresses/sources and automatic updates of the data which holds these

duncanbarnes:
From Settings > Devices > Edit Sources there is now the option to select the source from a drop down list populated from the source. The option to enter manually is retained. This will currently only work for OBS and in theory for Tricaster, the storage objects used for ATEM and Analog Way appear to be different abstraction objects, perhaps worth harmonising in the future?

OBS Source tally list is updated automatically when the OBS instance adds, renames or deletes. Renames are rippled through, deletes are not.

### DIFF
--- a/index.js
+++ b/index.js
@@ -398,6 +398,38 @@ function initialSetup() {
 		//gets all Tally Device Sources
 		res.send(device_sources);
 	});
+    
+    appSettings.get('/source_tallydata/:sourceid', function (req, res) {
+		//gets all Source Tally Data
+        let sourceId = req.params.sourceid;
+        let source = sources.find(src => src.id === req.params.sourceid);
+        let result = false;
+        
+        //Note: results from all except OBS and Tricaster commented out as either no device currently in warehouse to test with and/or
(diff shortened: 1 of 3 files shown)


josephdadams:
@duncanbarnes thank you! I'm a bit behind on some things but I will get to this when I can. Looks like a great addition!

duncanbarnes:
No problem, as a suggestion, it might be a good idea to split the different sections of code up into modules, 4500 lines is a challenging file to work with. It could be split up fairly easily with an MVC type structure but I've not had time to look through it fully.

josephdadams:
Agreed. It's very difficult at the moment. Any help you can add in this direction would be much appreciated.

duncanbarnes:
Am hoping to find some time, I hope you didn't take offence by the way, what you've built is an incredibly useful tool!

josephdadams:
None taken at all. I want this to be a tool for the community by the community, so I welcome any contributions. 😃

PR #56, opened 2021-02-14
Always send devices list when a M5Stick registers as a listener.

dakahl:
If the M5Stick was once connected to the server and one would change the name of a device, it does not get updated on the M5Stick because the deviceID is still valid.

Secondly, if there is a valid device ID the device_states are not getting send during connection and only on the next switch.

Both issues where fixed by moving the sending the information outside of the if statement.

Note: The devices message must be send before changing the deviceId to set the correct DeviceName that gets displayed on the M5Stick.


josephdadams:
Great job!

PR #58, opened 2021-02-21
Add batch file for easier startup.

dakahl:
Can be linked to the desktop an also to the autostart.


josephdadams:
I've decided to close this PR since it is windows-specific and not something all users will need; however, I think it is good material so I suggest adding it to the new wiki that has been established.

PR #60, opened 2021-02-27
Allow HTTPS JSON Requests for Outgoing Webhooks

br8kpoint:
I modified the outgoing web hook mechanism to extend it so that it can be used for sending json requests.

I personally use this to setup actions to control tp-link smart plugs through the outgoing web hook actions.

I added the protocol because the service communicates over https.

I also added content to set the content type of the request. JSON Will set the content-type to JSON. RAW will not set the content-type.

image

### DIFF
--- a/index.js
+++ b/index.js
@@ -233,10 +233,12 @@ var output_types_datafields = [ //data fields for the outgoing actions
 		]
 	},
 	{ outputTypeId: 'ffe2b0b6', fields: [ //Outgoing Webhook
-			{ fieldName: 'ip', fieldLabel: 'IP Address', fieldType: 'text' },
+			{ fieldName: 'protocol', fieldLabel: 'Protocol', fieldType: 'dropdown', options: [ { id: 'http://', label: 'HTTP' }, { id: 'https://', label: 'HTTPS'} ] },
+			{ fieldName: 'ip', fieldLabel: 'Ip/URL', fieldType: 'text' },
 			{ fieldName: 'port', fieldLabel: 'Port', fieldType: 'port' },
 			{ fieldName: 'path', fieldLabel: 'Path', fieldType: 'text' },
 			{ fieldName: 'method', fieldLabel: 'Method', fieldType: 'dropdown', options: [ { id: 'GET', label: 'GET' }, { id: 'POST', label: 'POST'} ] },
+			{ fieldName: 'content', fieldLabel: 'Content', fieldType: 'dropdown', options: [ { id: 'JSON', label: 'JSON' }, { id: 'RAW', label: 'RAW'} ] },
 			{ fieldName: 'postdata', fieldLabel: 'POST Data', fieldType: 'text' }
 		]
 	},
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -245,10 +245,12 @@ var output_types_datafields = [ //data fields for the outgoing actions
 		]
 	},
 	{ outputTypeId: 'ffe2b0b6', fields: [ //Outgoing Webhook
-			{ fieldName: 'ip', fieldLabel: 'IP Address', fieldType: 'text' },
+			{ fieldName: 'protocol', fieldLabel: 'Protocol', fieldType: 'dropdown', options: [ { id: 'http://', label: 'HTTP' }, { id: 'https://', label: 'HTTPS'} ] },
+			{ fieldName: 'ip', fieldLabel: 'Ip/URL', fieldType: 'text' },
 			{ fieldName: 'port', fieldLabel: 'Port', fieldType: 'port' },
 			{ fieldName: 'path', fieldLabel: 'Path', fieldType: 'text' },
 			{ fieldName: 'method', fieldLabel: 'Method', fieldType: 'dropdown', options: [ { id: 'GET', label: 'GET' }, { id: 'POST', label: 'POST'} ] },
+			{ fieldName: 'content', fieldLabel: 'Content', fieldType: 'dropdown', options: [ { id: 'JSON', label: 'JSON' }, { id: 'RAW', label: 'RAW'} ] },
```
josephdadams, line 253: Instead of these names (raw, json) can we use the official content types?

`application/json
`application/x-www-form-urlencoded`
`application/xml`
`text/plain`

And name the field “Content-Type” so it’s clear for all?

josephdadams (commented):
Left some code line feedback. Thanks! Great work!

index.js
```diff
@@ -4178,21 +4180,29 @@ function RunAction_TSL_31_TCP(data) {
...
+			options.headers['Content-Type']= 'application/json';
+
+		}
 		if (data.method === 'POST') {
 			if (data.postdata !== '') {
 				options.data = data.postdata;
 			}
 		}
-
+		logger(JSON.stringify(options), 'quiet')
```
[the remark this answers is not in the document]
br8kpoint: does info-quiet put it in the log box in the ui? I was thinking that type of data could just go in the console for debugging if the actions are not working, At least you would see what was being sent for troubleshooting. I found it helpful especially when doing the initial setup of new actions. Let me know what you think. I'm good either way. I do like to see it though somewhere.

index.js
br8kpoint: I can make this modification. what would we want the default to be? Would we want to have an option where the content type is not set?

index.js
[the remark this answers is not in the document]
josephdadams: `info-quiet` logs to console only

`info` logs to the console and the ui

`quiet` is not a type

index.js
josephdadams: I’m sure the axios library has a default, whether we set it or not. I would say let the default be `application/json` for now.

br8kpoint: I have added a Default option in the list which does not set the parameter. It will just set it to the axios default of url form encoded.

br8kpoint:
I have fixed the two issues in the pull request and tested the changes. They are good suggestions. Thanks.

josephdadams:
Great job! Thanks!

josephdadams:
I did go make a couple of tweaks to it - the most important being renaming the `.content` to `.contentType`, I felt like this was best for long-term clarity. As you were likely the only user of the `.content`, before, you may want to fix or delete any existing webhooks using that so that the new property will be available.

br8kpoint:
sounds good. I'll merge that back into my version. BTW. I made a tutorial on how we use Tally Arbiter with the Outgoing Webhook modification to control lights hooked up to TP-Link Kasa smart plugs to show active cameras. https://www.youtube.com/watch?v=TEAo6f4RJIE&t=4s

josephdadams:
Great! I like this as a solution because it makes the tally devices themselves super simple. You can literally just plug in a lamp. Mind if I link it?

josephdadams:
By the way, you can send TCP directly to the kasa plug to turn it on or off without having to do all the work to build the URL and go through tp-link's server. Just send it `0000002AD0F281F88BFF9AF7D5EF94B6C5A0D48BF99CF091E8B7C4B0D1A5C0E2D8A381F286E793F6D4EEDFA2DFA2` followed by CRLF and it will turn on.

`0000002AD0F281F88BFF9AF7D5EF94B6C5A0D48BF99CF091E8B7C4B0D1A5C0E2D8A381F286E793F6D4EEDEA3DEA3` and CRLF and it will turn back off.

We have a ton of these plugs all around our facility and turn them on and off all day like this. :)

br8kpoint:
sure go ahead and link it. nice to know about the direct tcp way too. always nice to have options.

br8kpoint:
FYI I also made another tutorial on my experience using the m5 sticks with Tally Arbiter Server with a couple of extra tips on getting it to work. https://www.youtube.com/watch?v=44NTLWL48L0

PR #85, opened 2021-05-14
Added Support for AV-HS410 Video Mixers

McHauge:
Hi, added support for Panasonic AV-HS410 mixers. 
I also added some yarn scripts for simpler dev work and deamon setup.
and a parameter for choosing what port to launch the vMix emulator on, just for those cases where Tally Arbiter is used on a system with vMix running.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,6 +22,8 @@ const ioClient		= require('socket.io-client');
 const osc 			= require('osc');
 const xml2js		= require('xml2js');
 const jspack 		= require('jspack').jspack;
+const os 			= require('os') // For getting available Network interfaces on host device
+
 
 //Tally Arbiter variables
 const listenPort 	= process.env.PORT || 4455;
@@ -38,6 +40,7 @@ const socketupdates_Settings  = ['sources', 'devices', 'device_sources', 'device
 const socketupdates_Producer  = ['sources', 'devices', 'device_sources', 'device_states', 'listener_clients'];
 const socketupdates_Companion = ['sources', 'devices', 'device_sources', 'device_states', 'listener_clients', 'tsl_clients', 'cloud_destinations'];
 const oscPort 		= 5958;
+const vmixEmulatorPort = '8099'; // Default 8099 
 var oscUDP			= null;
 var vmix_emulator	= null; //TCP server for VMix Emulator
 var vmix_clients 	= []; //Clients currently connected to the VMix Emulator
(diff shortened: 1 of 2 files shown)


alexcaglio83:
Good morning
i'm new on github.
i believe port 60020 is incorrect for panasonic av-hs410.
the default port is 60040.
sorry I'm not a programmer
I attach image link error.
Cattura

McHauge:
@alexcaglio83 Hi, that is both correct and incorrect, for this to work you need both the AUX_IP and the HS410_IF module to be installed on the mixer itself. and you will then have port 60040 for control only, and port 60020 provides both control and feedback with multicast support, therefore port 60020 is the correct one to use when we are talking Tally data :)


hope this helps

McHauge:
Basically I'm using the protocol made for this device [LINK](http://www.venetex.jp/venetexproduct/vs-r45/?lang=en) in order to get tally data, as there is no way to get it from the normal control interface plugin for the HS410

Here is the documentation for the protocol:
[AV-HS410-AUXP_IP-Protocol-Ver1.3EVol_2.pdf](https://github.com/josephdadams/TallyArbiter/files/6479753/AV-HS410-AUXP_IP-Protocol-Ver1.3EVol_2.pdf)

alexcaglio83:
thank you very much for helping.
soon I will test the system

udoschneider:
> Basically I'm using the protocol made for this device [LINK](http://www.venetex.jp/venetexproduct/vs-r45/?lang=en) in order to get tally data, as there is no way to get it from the normal control interface plugin for the HS410

Just out of curiosity: Wouldn't sending a `QBST` Request with Parameter `02` (PGM) or `03` (PVW) provide the appropriate response? From reading the docs I'd assume that the `ABST` Response would contain the active PGM/PVW crosspoint in it's second parameter.

McHauge:
@udoschneider where did you find a QBST command for the HS410, i must have missed it while reading though? also that would requerer pooling the unit all the time, vs it just  sending the data out to us whenever a change is detected.

udoschneider:
> @udoschneider where did you find a QBST command for the HS410, i must have missed it while reading though? also that would requerer pooling the unit all the time, vs it just sending the data out to us whenever a change is detected.

Checkout [AV-HS410-HS410_IF-Protocol-Ver1.3EVol_1.pdf](https://eww.pass.panasonic.co.jp/pro-av/support/dload/hs410_if201/AV-HS410-HS410_IF-Protocol-Ver1.3EVol_1.pdf), page 10.

The obvious drawback of course is the fact that you'd have to poll the switcher on a regular basis. One advantage would be that the same protocol is spoken by the HS-50 Switcher (which doesn't support TSL at all): See [AW-HS50InterfaceSpecifications-V1.00E.pdf](https://eww.pass.panasonic.co.jp/pro-av/support/content/guide/DEF/HS50_IP/AW-HS50InterfaceSpecifications-V1.00E.pdf), page 8.

I'm also curious if this really is the only tally protocol the HS-50 is using. When used in conjunction with the RP-50 PTZ Controller the RP-50 controls the tally of connected AW-HE cameras. This means that the RP-50 has to either poll the HS-50 as well or ... there is sth. undocumented going on. I may have to trace the traffic between those two maybe ....

I also noticed that the `ABST` response has a dedicated tally parameter. Not quite sure if this is redundant when querying PGM/PVW busses.

McHauge:
@udoschneider I totally missed that command while reading the dock, I'll look into adding support for that as well and then have a tick box for enabling pooling and one for multicast or something like that. but currently, I'm booked full so that would be at a later point. also, I'll properly add support for it in Bitfocus Companion first, as I'm more familiar with that process, just like I did with the current level of support.

PR #88, opened 2021-05-23
refactor for source types

josephdadams:
A beginning of the source types refactoring into their own files/classes


josephdadams:
@hrueger this is what I had started on before the ts conversion

josephdadams:
I will close this as it's outdated now.

PR #87, opened 2021-05-22
Frontend rewrite using Angular

hrueger:
**Advantages**:
* HTML elements are not created in strings and then added to the DOM, instead Angular handles all of that based on the template and its logic
* Removed the call to the external QR Code Server as it is now generated in the frontend
* No code duplication (e.g. navbar)
* improved credentials form
* Listener Clients in Producer View are sorted based on their Active / Inactive Status
* Complete typings for all Objects
* Improved UI with different colored buttons and icons
* Loading Spinners
* Progressive Web App functionality (e.g. to Home Screen)
* better confirm dialogs
* Chat History is kept until page refresh

**Current Status**:
- [x] Base structure
- [x] Home Screen
- [x] Tally Screen
- [x] Producer Screen
- [x] Settings Screen
- [x] Cloud Functionality
- [x] Icons
- [x] Fix no device states on initial load (producer screen)
- [x] Move sockets into service
- [x] SweetAlert
- [x] Loading indicators
- [x] Separate About Page
- [x] Redesign Home Page
- [x] Ports in use
- [x] Validation
- [x] Authentication
- [x] Chat
- [x] NoSleep
- [x] Favicon
- [x] PWA Stuff
- [x] Build Process
	- [x] CLI
	- [x] Electron
- [x] GitHub Actions

fixes #86

PR #94, opened 2021-06-15
TMP fix for running in subdirectory (and with reverse proxies)

MatteoGheza:
This is only a tmp "hack" while waiting for #87


josephdadams:
@MatteoGheza I've merged #87, so I'll close this. If you disagree, let me know.

Issue #20, opened 2020-11-27
Blackmagic ATEM - Show PVW/PGM status when source is used in Upstream Key


aacoba:
ATEM supports using an input as an "picture in picture" overlay using the UpstreamKey (USK) DVE functionalities.
Currently TallyArbiter does not recognize an input as being "used" in the USK and does not update the tallies accordingly.
It looks like the PGM state of the USK can be found in https://nrkno.github.io/tv-automation-atem-connection/interfaces/upstreamkeyer.html by looking at `onAir` and `fillSource`.

The PVW state looks to be a bit more difficult to determine. When I include the key in the next transition (include it in the PVW) I see that the `video.ME.0.transitionProperties` state gets updated.
The content of the state is `{ style: 0, selection: 3, nextStyle: 0, nextSelection: 3 }` where both selection and nextSelection change with the keys included in the next transition. 
It looks like the value of `selection` and `nextSelection` can be translated back into the binary version of which keys are included in the transition.
![image](https://user-images.githubusercontent.com/659308/100484884-6dfc2680-30fe-11eb-8db6-55f1b39609b0.png)
For example:
Only BKGD -> 1
BKGD + KEY1 -> 3
KEY1 -> 2

It would be very useful for us if TallyArbiter is able to represent this correctly in the future!

josephdadams:
@aacoba I will see if I can get this added at some point!

wicherdt:
Yes, it is very important in green screen virtual productions.
Please make it.

ruebyi:
Would be great to see this in a future Update - the new ATEM Extreme has 4 Upstream Keyers, so it would be very helpful to get feedback about their Status.

josephdadams:
This should be fixed in latest

Issue #50, opened 2021-01-30
ATEM tally incorrect during a transition


smadds:
This is such a great project!. I love it!

If I start a transition manually on an ATEM Studio HD the PV bank switches to red, but this is not matcher on TA. It only seems to notice at the end of the transition, when the new input hits the program bus. The screenshot shows the 2 reds on the ATEM, but the setup mimic on TA shows a red & green. Is this a problem with the signalling from the ATEM?

FYI my Stream Deck Companion status works fine when linking the button's status to "Change colours from mixer program tally"

 
![image](https://user-images.githubusercontent.com/5259870/106368877-0b651800-6345-11eb-8f85-d01320936916.png)

jerriley:
I noticed this as well. Even after the Mix transition is complete, there is a delay of a second or two before the colors swap in Tally Arbiter (and on the listeners I use). The bug is the same for Auto button transition as well as T-bar transition.

josephdadams:
This should be fixed in latest

PR #95, opened 2021-06-21
ATEM: use tally logic implemented in atem-connection library

peschuster:
Rewrite of processing of ATEM changes to use `listVisibleInputs` function implemented in atem-connection library.

Features:
- during transition both sources (preview and program) are red (i.e. on program)
- also handles sources in keyers, downstream keys and super source

Fixes issues #20 and #50

I hope I didn't break anything, but as far as I could see all changes should be local to a single atem source.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,28 +1,29 @@
 /* Tally Arbiter */
 
 //Protocol, Network, Socket, Server libraries/variables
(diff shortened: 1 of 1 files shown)


josephdadams:
Great job!

PR #103, opened 2021-07-08
Initial Docker and Gitpod support

MatteoGheza:
@hrueger https://github.com/josephdadams/TallyArbiter/pull/87#issuecomment-846372039 if you want to try this, I suggest you to go to https://gitpod.io/#https://github.com/josephdadams/TallyArbiter/pull/103 and login with Github

### DIFF
--- a/docs/docs/installation/docker.md
+++ b/docs/docs/installation/docker.md
@@ -10,6 +10,18 @@ You can pull the image from DockerHub using the following command:
 docker pull tallyarbiter:latest
 ```
 
-Then, start it by typing `docker run -p 4455:4455 -v $(pwd)/config.json:/app/config.json tallyarbiter`.
+Then, start it by typing `docker run -p 4455:4455 -v $(pwd)/config.json:/app/config.json --restart unless-stopped tallyarbiter`.
+If you preferr using docker-compose, you can use this configuration:
+```yaml
+version: '3.3'
+services:
+    tallyarbiter:
+        ports:
+            - '4455:4455'
+        volumes:
+            - '$(pwd)/config.json:/app/config.json'
+        restart: unless-stopped
+        image: tallyarbiter
+```
 
 **Be sure to mount your [config file](#configuration) using Docker volumes!**
--- a/docs/docs/installation/docker.md
+++ b/docs/docs/installation/docker.md
@@ -4,13 +4,13 @@ sidebar_position: 2
 
 # Docker
 ## Installation
-If you have Docker installed, you can install TallyArbiter using our official Docker image [`insert Docker image name`](https://hub.docker.com/r/matteogheza/tallyarbitertestimage).
+If you have Docker installed, you can install TallyArbiter using our official Docker image [`TallyArbiter`](https://hub.docker.com/r/josephdadams/tallyarbiter).
 You can pull the image from DockerHub using the following command:
 ```bash
-docker pull dockerimagename:latest
+docker pull tallyarbiter:latest
 ```
 
-Then, start it by typing `docker run -p 4455:4455 dockerimagename`.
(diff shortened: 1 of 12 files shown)


hrueger:
Looks good to me 👍
Maybe you can add a GitHub Action Workflow which runs on pushes and builds the image and if it is a tag, publishes the image to Docker Hub?

MatteoGheza:
Sorry, I didn't read the comment (I was trying to setup a simple e2e testing system using cypress.io, and I had some problems with a library used for intercepting WebSockets, so I had to fork it).
Tomorrow i'm going to try.
Thanks for the feedback.

josephdadams:
Also add docker and gitpod instructions somewhere in the readme, docusaurus, etc. so people can know about it :)

MatteoGheza:
GitHub Action Workflow it's working.
https://hub.docker.com/r/matteogheza/tallyarbitertestimage/tags?page=1&ordering=last_updated
https://github.com/MatteoGheza/TallyArbiter/runs/3027417377?check_suite_focus=true

MatteoGheza:
I started adding references to Docker and Gitpod, but I'm not good at editing documentation :Joy:

hrueger:
@MatteoGheza 
>Fix GitHub Actions Workflow (Yaml syntax error, @hrueger help wanted 😅)

If I remove all `if: !(startsWith(github.ref, 'refs/tags/v'))`, the yaml becomes valid. Maybe we need to escape the `!` somehow? I'm not that familiar with yaml.

hrueger:
According to https://github.community/t/expression-syntax-for-not-startswith/17040/4 we need to put the expression in double quotes or as stated in https://github.community/t/expression-syntax-for-not-startswith/17040/3 the following should work:
```yml
if: startsWith(github.ref, 'refs/tags/v') !== true
```

I'd personally prefer double quotes.

MatteoGheza:
Now it's working! :hurray:

josephdadams:
I added the repo secrets, modified the yml for the correct docker path, and updated the documentation to include that path as well.

josephdadams:
@MatteoGheza ready to merge?

MatteoGheza:
I found a small TODO in documentation.
Sorry for the late reply.

docs/docs/installation/docker.md
```diff
@@ -0,0 +1,16 @@
...
+# Docker
+## Installation
+If you have Docker installed, you can install TallyArbiter using our official Docker image [`TallyArbiter`](https://hub.docker.com/r/josephdadams/tallyarbiter).
+You can pull the image from DockerHub using the following command:
+```bash
+docker pull tallyarbiter:latest
+```
+
+Then, start it by typing `docker run -p 4455:4455 tallyarbiter`.
+TODO: explain mounting config.json
```
MatteoGheza, line 14: We need to explain how to mount config.json to a file outside Docker container

josephdadams: I don't use Docker at all so I have no idea how to do this.

hrueger: Let's provide a `docker run` command as well as a `docker-compose` example. I can add that tomorrow.

MatteoGheza: Partially done. Need testing, because in the current env i can't use Docker.

hrueger: I don't think that $(pwd) works on Windows but I can check that later today. If it does not work, we could use an absolute path there so that the command stays cross-platform compatible?

MatteoGheza:
Any updates? Has anyone tried this?

hrueger:
Works for me 👍

josephdadams:
Merged. If anyone has an issue we can troubleshoot. 😂

PR #121, opened 2021-07-27
Limit UI to 5000 lines of logs and 5000 lines of TallyData and delete first log line after reaching that point

JTF4:
With the current solution the logs and tallydata objects get to big for web browsers to load. I limited the logs and tally data in the UI to 10000 lines to hopefully fix this issue. After reaching the limit, the object it purges and it starts with an empty array.


josephdadams:
@JTF4 what would it take to limit it to the most recent 1,000 or 5,000 instead of deleting the entire log every 10,000? Just letting it rise to the threshold and then never getting bigger, but deleting the oldest entry every time a new one is added?

JTF4:
Yes. I will make that update tomorrow.

josephdadams:
We should also write the tally data to a log file. Maybe a new file every day, or run an interval and if no tally data is received after x minutes, start a new file next time. It could be used to generate EDL data perhaps.

JTF4:
@josephdadams I have mode the edit. 

New behavior is after 5000 lines generated, the first line is removed once a new line is generated.

JTF4:
Also, I will start working on the log file but I will file that under a new PR

PR #127, opened 2021-07-29
Writes logs and tally data to their own file and deletes any log or tally data file older than 1 week

JTF4:
The logs are written to a logs folder under the TallyArbiter folder and the tally_data is written to a TallyData folder in the same place 

The logs use the .talog extension and tally data uses the .tadata extension. 

I am using the find-remove package to find and remove log and tally data files older than 1 week.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1914,7 +1914,7 @@ function logger(log, type) { //logs the item to the console, to the log array, a
 	logObj.type = type;
 	Logs.push(logObj);
 
-	writeLogFile(dtNow, log);
+	writeLogFile(log);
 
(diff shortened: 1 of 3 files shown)


hrueger:
Maybe instead of writing the whole logs to a file every time a line is added, we could just append that line to the file. And instead of opening and closing it, a stream could be used. See https://stackoverflow.com/a/43370201/13485777

What do you think?

JTF4:
Sounds like a good idea to me! Super quick change too.

JTF4:
Done!

PR #135, opened 2021-08-02
Initial e2e testing system support

MatteoGheza:


UI/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
hrueger, line 1: Do we need that?

hrueger (approved):
Looks great!

UI/cypress/fixtures/example.json
MatteoGheza: No, it was generated adding Cypress schematic to Angular.

hrueger: 👍

MatteoGheza:
@hrueger @josephdadams @JTF4 Now everything is working, but I'll be Away From Keyboard until Wednesday 25, so I can't contribute to TallyArbiter in those days.
In the next week I'll finish E2E tests.

TODOs:
- [ ] Add tests for every settings section, for example "Devices", "Sources" and "Cloud"
- [ ] Run a test without intercepting and simulating websockets, so we can test backend. This is going to be used only before releases or pull requests, not for every commit
- [ ] Decide if we're going to use a free account on cypress.io for recording tests in a dashboard and for running test in parallel.  
BTW it's free, but there are limitations until you verify your account as an "account related to an Open Source project".  
For more information: https://www.cypress.io/pricing https://docs.cypress.io/guides/dashboard/organizations#Open-Source-Plan
Anyway, I don't think we need to view tests in a custom dashboard or to run them in parallel, since Github Actions machines are pretty fast.

PR #138, opened 2021-08-02
Osc sourcetype patch

JTF4:
When you try to create an OSC source type, it will throw an error because it think that the information field needs to be filled.


josephdadams:
Looks good. Is the try catch revert just something else you are working on?

JTF4:
Yes. It was until I realized that it was an issue on my end with a config file. After deleting my config and trying again it stopped happening.

JTF4:
@josephdadams are you ok if I go ahead and deploy this?

josephdadams:
Sure

PR #139, opened 2021-08-03
Support for rate limiter and brute force protection

MatteoGheza:
Currently, logins are vulnerable to brute force, since there are no rate limits to the login event in websocket.
I tried implementing rate limits using an advaced library, [rate-limiter-flexible](https://github.com/animir/node-rate-limiter-flexible), so, in the future, we can implement rate limits to the entire application using a custom Express middleware (for example [https://github.com/animir/node-rate-limiter-flexible/wiki/Express-Middleware](https://github.com/animir/node-rate-limiter-flexible/wiki/Express-Middleware)) and/or Websocket connections (https://github.com/animir/node-rate-limiter-flexible/wiki/Overall-example#websocket-single-connection-prevent-flooding).
P.S: I added "-1" in `login_response` to maintain compatibility with "old `login_response` versions"

### DIFF
--- a/UI/src/app/_components/login/login.component.html
+++ b/UI/src/app/_components/login/login.component.html
@@ -1,7 +1,7 @@
 <div class="container text-center d-flex justify-content-center mt-5 pt-5">
     <main class="form-signin">
         <h1 class="h3 mb-3 fw-normal">Please sign in</h1>
-        <div class="my-2 text-danger" *ngIf="loginFailed">{{ errorMessage }}</div>
+        <div class="my-2 text-danger" *ngIf="!loginResponse.loginOk">{{ loginResponse.message }}</div>
         <div class="form-floating">
             <input type="text" class="form-control" (keydown.enter)="inputPassword.focus()" [(ngModel)]="username" id="username" placeholder="Username">
             <label for="username">Username</label>
--- a/UI/src/app/_components/login/login.component.html
+++ b/UI/src/app/_components/login/login.component.html
@@ -1,7 +1,7 @@
 <div class="container text-center d-flex justify-content-center mt-5 pt-5">
     <main class="form-signin">
         <h1 class="h3 mb-3 fw-normal">Please sign in</h1>
-        <div class="my-2 text-danger" *ngIf="wrongUsernameOrPassword">Wrong username or password!</div>
+        <div class="my-2 text-danger" *ngIf="loginFailed">{{ errorMessage }}</div>
         <div class="form-floating">
             <input type="text" class="form-control" (keydown.enter)="inputPassword.focus()" [(ngModel)]="username" id="username" placeholder="Username">
             <label for="username">Username</label>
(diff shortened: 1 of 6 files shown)


JTF4:
I could break into this much faster without brute force. 

Since the entire thing is running with a non https socket you could easily sniff out the password during the username and password exchange. 

But I do like the rate limiting.

hrueger (changes requested):
Cool.
>P.S: I added "-1" in login_response to maintain compatibility with "old login_response versions"

Just wondering: compatibility with what? 

I personally would like to see the error message generated on the server side. That way we could implement more errors (like from a user access control with specific rights for specific users).
What do you think?

MatteoGheza:
@hrueger sorry for the late response

> Cool.
> 
> > P.S: I added "-1" in login_response to maintain compatibility with "old login_response versions"
> 
> Just wondering: compatibility with what?

I meant compatibility with older versions of the UI.
If we change "login_response" response from boolean to an object (for example,  
`{ "status":"success", "message":"" }`  
`{ "status":"error", "message":"Wrong username or password!" }`  
`{ "status":"error", "message":"Too many attemps! Please retry later." }`  
) the old version of `auth.service.ts` stops working. 

> I personally would like to see the error message generated on the server side. That way we could implement more errors (like from a user access control with specific rights for specific users).

So do I.

> What do you think?

Using server-side generated error messages, we could also transmit to the client the number of attempts left and how long it has to wait before a new attemp.
I think we can use a new websocket event, for example `login_response`, so the server can emit both `login_result` with the old "boolean" response and the new `login_response` with an object.

hrueger:
>sorry for the late response
No problem.

> meant compatibility with older versions of the UI.
I might be wrong, but why would we need that? The only clients having that UI are the browser clients and when you refresh them, the latest UI is being fetched, or is it not?

hrueger:
Or do you mean caching from the service worker? That could indeed be a problem. However, I'm not a fan of legacy code ;-) My suggestion would be to display a changelog message on every first start of a new version. We could save the current version in the config file and once it does not match anymore (because TA was updated), we show the changelog.
In there we could inform the user to refresh the browser to get the latest UI.

MatteoGheza:
> I might be wrong, but why would we need that? The only clients having that UI are the browser clients and when you refresh them, the latest UI is being fetched, or is it not?

I meant that probably some people who are using Tally Arbiter from source don't always remember to re-compile the UI when they pull from Git

MatteoGheza:
> However, I'm not a fan of legacy code ;-)

I think we can "deprecate" login_result with boolean in the next "big release"

hrueger:
Ok, I agree

hrueger:
>I meant that probably some people who are using Tally Arbiter from source don't always remember to re-compile the UI when they pull from Git.
We could add a proxy for `localhost:4200` so that when TallyArbiter is started in dev mode (NODE_ENV=production) not the bundled UI is served but instead `localhost:4200`.

MatteoGheza:
I think that now login rate limits are working as requested

PR #143, opened 2021-08-04
Removing line 949 and extra lines

JTF4:


JTF4:
Tested and seems to patch the issue with OBS

MatteoGheza (approved):

PR #145, opened 2021-08-04
Add Atem cut bus support

JTF4:


JTF4:
Super simple function. If in cut bus mode, then TA will ignore all preview states from the ATEM, but if it's in preview / program mode, then all states will be reflected.

josephdadams:
Is there a way to detect what mode the ATEM is in? What if the user changes it after configuring the source?

JTF4:
If you go into the source settings, you can set what mode its on and whenever it receives a new tally data it will check what mode is configured. I will look at the AETM web socket to see if they report what mode it's in. ￼

JTF4:
It does not look like there is a way to detect this with the functions contained in the sdk. If you know of a way, I would be happy to include it.

josephdadams:
@julusian I summon you. Is it possible to detect when an ATEM is in cut bus mode?

Julusian:
Well, its not the atem that is in cut bus mode, just the built in panel. So it is still possible to use use the preview remotely.

I think it is possible but I believe it is done via a separate protocol (used only by the ATEM Setup app), and I am not aware of any work being done to change any settings there

JTF4:
Ah. Ok. So it seems that it would be pretty difficult to detect a ATEM mini in cut bus mode automatically as we don't have access to their setup protocol.

I would love the reverse engineer that in the future, but at the moment I don't have access to an ATEM mini to test the settings protocol on.

josephdadams:
@JTF4 you may merge when ready

PR #146, opened 2021-08-04
Electron improvements

MatteoGheza:
- Ignore files for packing app.asar and add category
- Fix documentation opening in the same window


hrueger:
Lgtm. If we want to reduce the bundle size quite a bit more, we could add something like vercel's ncc, as this would produce a single js file and make the node_modules folder completely obsolete. However, I don't know if it works together with electron.

PR #152, opened 2021-08-07
Filter sourceTypes by enabled

### LINKED ISSUE #150: Analog Way Livecore unable to connect.
Unable to key in IP address and port number for analog way live core.

### Version

2.0.8

Safari

### Relevant log output

_No response_

MatteoGheza:
Fix #150


hrueger:
Thanks. Now we could remove the filter function from the settings component's template as it is unneccessary.

PR #176, opened 2021-08-22
bus option color support

josephdadams:
Adds support to edit bus option colors; the listener clients will have to be modified to support these colors as most of them are currently hard coded to green/red

### DIFF
--- a/UI/src/app/_components/settings/settings.component.html
+++ b/UI/src/app/_components/settings/settings.component.html
@@ -291,7 +291,7 @@
               <tr *ngFor="let busOption of socketService.busOptions">
                 <td>{{busOption.label}}</td>
                 <td>{{busOption.type}}</td>
-                <td>{{busOption.color}}</td>
+                <td><span class="color-box me-3" [style.background-color]="busOption.color"></span>{{busOption.color}}</td>
                 <td>
                     <button class="btn btn-outline-dark" (click)="editBusOption(busOption, busOptionModal)">
                         <i class="fas fa-pen"></i> Edit
--- a/UI/src/app/_components/settings/settings.component.html
+++ b/UI/src/app/_components/settings/settings.component.html
@@ -1,189 +1,184 @@
 <div class="container pt-4" *ngIf="socketService.initialDataLoaded">
-  <div class="row">
-    <h2>
-      Sources
-      <button class="btn btn-outline-success me-2" (click)="addSource(sourceModal)">
+    <div class="row">
+        <h2>
+            Sources
+            <button class="btn btn-outline-success me-2" (click)="addSource(sourceModal)">
         <i class="fas fa-plus"></i> Add 
       </button>
-      <button class="btn btn-outline-dark" (click)="toggleTestMode()">
+            <button class="btn btn-outline-dark" (click)="toggleTestMode()">
         <i class="fas fa-vial"></i> Turn {{ socketService.testModeOn ? "Off" : "On" }} Test Mode
       </button>
-    </h2>
-    <span class="text-muted mt-3" *ngIf="socketService.sources.length == 0">No sources configured.</span>
-    <table class="table table-hover" *ngIf="socketService.sources.length > 0">
-      <thead>
-        <tr>
-          <th>Name</th>
-          <th></th>
-          <th>Type</th>
-          <th></th>
-        </tr>
-      </thead>
-      <tbody>
-        <tr *ngFor="let source of socketService.sources" [class]="
+        </h2>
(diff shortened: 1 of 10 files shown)


josephdadams:
@hrueger can you help implement this in the web tally viewer? In the bus_options array, there is now a `color` property.

hrueger:
Why did you comment out the `previewprogram` bus option?

josephdadams:
It's been commented out since possibly 1.0. Basically decided to implement the previewprogram state in the listener client level. But I am open to re adding it.

hrueger:
OK. I suggest re-adding it for now so that we can have people modify the yellow color, too,

josephdadams:
I am also giving some thought toward adding other aux busses as well and letting the user assign the incoming source type to what the TA bus is. For example, with TSL 3.1, they could say "tally1" should go to TA Preview.

josephdadams:
Each listener client will have to be modified to support this. Right now, it is up to the listener client to evaluate whether it's in preview, program, or both, and then what to do with that info.

josephdadams:
@hrueger if you have the skill to add a colorpicker for editing, please do. I will have to google my way through that

josephdadams:
nevermind, you already did :) :)

hrueger:
In the web tally the program priority is now the default mode and can be disabled by adding `?programPriority=false` to the url. I also added a paragraph to the docs. 

Could someone verify that this actually works? I don't have anything right now to output some tally data and the test mode does not (yet? 😉) provide the `programpreview` bus.

josephdadams:
I thought the plan was to remove the previewprogram bus as it is not really a bus but a reflection of the device being in multiple other busses.

hrueger:
Ah, sorry, I completely forgot about that ;-) So how exactly should we handle that in the web tally? We're using `previewprogram` there only to get the user's custom color from the settings.

josephdadams:
@hrueger take a look at the python script for the blink(1) client. Basically you should check the device's states (`device_states`) and display the color of the one that has the highest priority number. Then I thought it would be nice to display the other states in a small table on the screen as well.

```python
def processTallyData():
	busses = []

	for device_state in device_states:
		if len(device_state['sources']) > 0:
			bus = getBusById(device_state['busId'])
			busses.append(bus)
	
	priority = 0
	tallyBus = []

	for bus in busses:
		if bus['priority'] > priority:
			priority = bus['priority']
			tallyBus = bus
	
	#this should leave us with a tallyBus object where the highest priority bus is in this object
	evaluateMode(tallyBus)

def evaluateMode(bus):
	global deviceId
	device = getDeviceById(deviceId)
	if len(bus) > 0:
		deviceName = device['name']
		print('{} is in: {}'.format(deviceName,bus['type']))
		color = hexToRGB(bus['color'])
		doBlink(color[0], color[1], color[2])
```

hrueger:
Ah, that makes sense. Sorry that I didn't get that in the first place.
We should also modify the section in the settings where the devices are listed (to for example also show the aux outputs there).

However, with your latest commit my TA does not start anymore:
```
C:\Users\Hannes\Desktop\Angular\TallyArbiter\index.js:2376
        deviceStateObj.deviceId = deviceObj.id;
                                  ^

ReferenceError: deviceObj is not defined
    at initializeDeviceStates (C:\Users\Hannes\Desktop\Angular\TallyArbiter\index.js:2376:35)
    at loadConfig (C:\Users\Hannes\Desktop\Angular\TallyArbiter\index.js:2307:5)
    at startUp (C:\Users\Hannes\Desktop\Angular\TallyArbiter\index.js:816:5)
    at Object.<anonymous> (C:\Users\Hannes\Desktop\Angular\TallyArbiter\index.js:7302:1)
```

josephdadams:
I’ll fix that.

josephdadams:
The long term plan, I hope, is to let the user define as many busses in TA as they want. Then, when defining Device-to-Source Bus Mapping, they will, for example, need to say “Address 1” PGM on my ATEM is TA Bus “Aux 1”. This is a big overhaul.

hrueger:
Ah, I had to delete my config file and re-add the devices. We probably need some kind of migration to add the needed properties so that not everybody has to delete their config? (linkedBusses)

josephdadams:
I believe the fix I pushed will prevent the error from happening without deleting the config, but we do need to change the UI to support this new feature to be able to link any bus.

hrueger:
I still get the error when it is trying to `devices[i].linkedBusses.push` and `devices[i].linkedBusses.splice` ....
working on the UI.

josephdadams:
@hrueger can we delete this branch now?

hrueger:
Yes, I think so 👍

PR #177, opened 2021-08-22
Fix autoupdate on MacOS

hrueger:
I noticed in another one of my Electron apps that the autoupdate on MacOS was not working. So I checked TallyArbiter and we have the same problem here. Apparently the updater needs the application in a `.zip` file.


MatteoGheza:
Great!

PR #183, opened 2021-08-27
external Address option

videojedi:
Add a new value to config file. Will display in home page as a QR code and extra link.

example
 "externalAddress":"http://0.0.0.0:4455/#/tally"


hrueger:
@MatteoGheza
> Now we need to implement an input or a modal for replacing the external address and a new websocket event for applying changes to it in the config.

What about a global "Settings" page where users can also change other config settings like passwords? I'd also really like to see a users system for multiple users with different login credentials.

MatteoGheza (approved):

MatteoGheza:
I think we can merge this, so @videojedi can start using this as an "hidden feature" until we implement a "global settings" page.

hrueger:
Maybe someone can delete the `.code-workspace` file? For me it appears to still be there

MatteoGheza:
> Maybe someone can delete the `.code-workspace` file? For me it appears to still be there

I'll do it in master

MatteoGheza:
Done

Issue #186, opened 2021-08-29
[Feature Request]: Tour upon first load


josephdadams:
### What feature would you like to see added to TallyArbiter?

It would be great to give the user a tour of the Settings page upon first load (or on demand) that walks them through each section, what it is for, and how to use it.

hrueger:
Great idea. I'd recommend using [`ngx-tour` with the `ng-bootstrap` flavor](https://ngx-tour.github.io/ngx-tour/#/ng-bootstrap).

josephdadams:
Something you can do?

PR #190, opened 2021-08-29
Tour

hrueger:
Adds the tour from #186 

Still ToDo:

- [ ] Decide when to start the tour (maybe if the config file does not yet exist?)
- [ ] Add all tour elements and texts
- [ ] Decide whether to do seperate tours (one per component) or not

About the 3rd point: The `ngx-tour` can automatically navigate between pages. However, we need to be logged in for the tour on the settings page to work correctly. Maybe we need a second tour instance there? Or should we prompt the user for the credentials before starting the tour? Or even try the default credentials? Any ideas?


hrueger:
Adding a tour box is really simple:
1. Add `tourAnchor="home.producer"` to any html element. Use `[component].[someDescriptiveElementIdentifier]` for the value of the attribute.
2. Add
```
      {
        anchorId: 'home.settings',
        title: 'Settings',
        content: 'ToDo',
        enableBackdrop: true,
      },
```
to the array in the `home.component.ts`. Use the same Anchor ID and provide your own title and content.

3. If you want to navigate to another component, add `route: "/settings",` (or any other url) to the step config.

josephdadams:
I would say if no sources are configured, prompt for the tour

josephdadams:
Closing, stale.

Issue #195, opened 2021-08-30
Can we use Winston for logging?


MatteoGheza:
In #182 @hrueger wrote that we can use `winston` for logging, and I think that it's a good idea.
PROS of using `winston`:
- Log file rotation
- Support for multiple "transports": not only console or files, but also Syslog, Windows Event Viewer or Amazon Cloudwatch

I think we can use two loggers: one, with "file" and "console" as transports, for logging, and one, with a file transport, for Tally data.
Tomorrow I'll try to open a PR, so we can decide if using `winston` is a good idea or not.

josephdadams:
Sounds great to me.

JTF4:
I like that too. I'm working on sentry error reporting as well. It will give us better insight into how everything is running and where we need to put out time in terms of bug fixing even when someone does not manually submit a bug report.

PR #197, opened 2021-08-31
Migrate logging and tally data to Winston

MatteoGheza:
As described in #195 
Everything is compatible with the old version of the logger, except that now every line of server logs is similar to
```[2021-08-31 21:53:19] info: Starting HTTP Server.``` or ```[2021-08-31 21:53:40] error: An error occurred etc```

-------------

_Winston options for file output:_
chrome_BNcSWGT34T

Current options used in server loggin and in Tally data:
```javascript
{
  filename: filepath,
  maxsize: 3e+7
  maxFiles: 3,
  ...
}
```

-------------

@JTF4 can you explain to me what is Tally data logging and why its format is
`{TallyObj},  {TallyObj}, {TallyObj},`?

### DIFF
--- a/index.js
+++ b/index.js
@@ -67,24 +67,24 @@ var Logs = []; //Used for loading logs in settings page
 var tallyDataFilePath = getTallyDataPath();
 
 const serverLoggerLevels = {
-  levels: {
-    critical: 0,
-    error: 2,
-    warning: 3,
-    console_action: 4,
-    info: 5,
-    'info-quiet': 6,
-    debug: 7
-  },
-  colors: {
-    critical: 'red',
-    error: 'red',
-    warning: 'yellow',
-    console_action: 'green',
-    info: 'white',
-    'info-quiet': 'white',
-    debug: 'blue'
-  }
+	levels: {
+	    critical: 0,
+	    error: 2,
+	    warning: 3,
+		console_action: 4,
+		info: 5,
+	    'info-quiet': 6,
+	    debug: 7
+	},
+	colors: {
+		critical: 'red',
+	    error: 'red',
+	    warning: 'yellow',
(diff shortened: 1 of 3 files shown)


JTF4:
@MatteoGheza the tally data logging is there so that we can enable edit decision list exporting in the future.

MatteoGheza:
> @MatteoGheza the tally data logging is there so that we can enable edit decision list exporting in the future.

@JTF4 can we change the format into
```javascript
[
  {TallyObj},
  {TallyObj},
  {TallyObj}
]
```
or into
```javascript
[
  { timestamp: current_timestamp, tally: {TallyObj} },
  { timestamp: current_timestamp, tally: {TallyObj} },
  { timestamp: current_timestamp, tally: {TallyObj} },
]
```

josephdadams:
I see no reason why we can't make that change. If we come up with an EDL feature in the future, I have some specific thoughts on how I would want to see it handled anyway, so it's fine with me to make changes to how it's currently logged to a file.

JTF4:
The latter should work. As long as the tally data object has the device name in it with the tally1-4 it will work. In the future I'll also add a time code to the log object so it can sync with a recorder like a ki pro.

MatteoGheza:
In future, when we start working on EDL and similar features, I think I can replace Winston with a custom function for storing Tally data in a better method then the one used now or then the method used before this PR.

josephdadams:
Sounds good. I shall open a new issue sometime to describe what I was thinking for EDL.

MatteoGheza:
@josephdadams @JTF4 is this ready? Do I have to revert edits to Tally Data logger?

josephdadams:
If it works (I haven’t had time to test), it’s fine with me to merge. W

PR #204, opened 2021-09-03
Typescript Refactoring

hrueger:
Hi all,
it has been more than two month since we started doing the Angular UI Rewrite, so it is time for another big refactoring ;-)
I'm trying to organize and restructure the big `index.js` file into multiple TypeScript classes.
It should be easier to maintain and add new features.

ToDo:
- [x] Initial TypeScript support
- [ ] Classes
- [ ] Sync models with UI instead of copying
- [ ] Get rid of `// @ts-ignore`s
- [ ] Get rid of all `any` types
- [ ] Add types for function parameters
- [ ] Add function return types
- [ ] Add Docs on how to add a Source or an Action (or sth else?)


hrueger:
Hm, actually I didn't want to start this branch based on the `bus-options` branch, but maybe this is not a bad idea as that one already has quite a few changes. @josephdadams should we continue bus options here?

josephdadams:
Fine with me to take the bus options branch this direction. 😃

josephdadams:
It would be great to have each source type’s logic and required libraries contained within itself or folder somehow. This would make it a lot more modular to add more source types and have them all emit data back to the main backend through a common structure. They sort of all do this already by evaluating their tally logic and then running the result through `processTSLTally`.

hrueger:
That is exactly what I'm working on. I thought that those classes would extend `EventEmitter` so they can send events "up" to the main application.

josephdadams:
I sort of started that in the refactor source types branch if it is helpful. Not TS though

hrueger:
@josephdadams This is what I came up with for the ATEM source. I did't use events but instead `rxjs`'s `BehaviourSubjects` which provide better types and can also provide the current state more easily.

I created some decorators for registering the source. No global if clause is needed anymore, because by using this decorator it is found automatically. 

Now I have a question for you: In what format should the source classes provide the tally information to the main application? Sorted by bus?

What if the Source can also provide (using a decorator) bus types which then all appear in the settings page?

```TypeScript

import { Atem, listVisibleInputs } from 'atem-connection';
import { logger } from '..';
import { RegisterSource } from "../_decorators/register-source.decorator";
import { UsesPort } from '../_decorators/uses-port.decorator';
import { Source } from '../_models/Source';
import { SourceClass } from './_Source';

@RegisterSource("44b8bc4f", "atem", "Uses Port 9910.", [
    { fieldName: 'ip', fieldLabel: 'IP Address', fieldType: 'text' },
    {
        fieldName: 'me_onair', fieldLabel: 'MEs to monitor', fieldType: 'multiselect',
        options: [
            { id: '1', label: 'ME 1' },
            { id: '2', label: 'ME 2' },
            { id: '3', label: 'ME 3' },
            { id: '4', label: 'ME 4' },
            { id: '5', label: 'ME 5' },
            { id: '6', label: 'ME 6' },
        ],
    },
    {
        fieldName: 'cut_bus_mode',
        fieldLabel: 'Bus Mode',
        fieldType: 'dropdown',
        options: [
            { id: 'off', label: 'Preview / Program Mode' },
            { id: 'on', label: 'Cut Bus Mode' }
        ],
    },
])
@UsesPort("9910")
export class ATEMSource extends SourceClass {
    private atemClient: Atem;
    constructor(source: Source) {
        super();


        let atemIP = source.data.ip;
        logger(`Source: ${source.name}  Creating ATEM Connection.`, 'info-quiet');
        this.atemClient = new Atem();

        this.atemClient.on('connected', () => {
            this.connected.next(true);
            logger(`Source: ${source.name} ATEM connection opened.`, 'info');
        });

        this.atemClient.on('disconnected', () => {
            this.connected.next(false);
            logger(`Source: ${source.name} ATEM connection closed.`, 'info');
        });

        this.atemClient.on('stateChanged', (state, path) => {
            for (let h = 0; h < path.length; h++) {
                if (path[h] === 'info.capabilities') {
                    //console.log(state.info.capabilities);
                }
                else if ((path[h].indexOf('video.mixEffects') > -1) || (path[h].indexOf('video.ME') > -1) || (path[h].indexOf('video.downstreamKeyers') > -1)) {
                    const pgmList = [], prvList = [];
                    const addUniqueInput = (n, list) => {
                        const s = n.toString();
                        if (!list.includes(s)) list.push(s);
                    }
                    for (let i = 0; i < state.video.mixEffects.length; i++) {
                        if (source.data.me_onair.includes((i + 1).toString())) {
                            listVisibleInputs("program", state, i).forEach(n => addUniqueInput(n, pgmList));
                            listVisibleInputs("preview", state, i).forEach(n => addUniqueInput(n, prvList));
                        }
                    }
                    this.tally.next({ pgmList, prvList });
                }
            }
        });

        this.atemClient.on('info', console.log);
        this.atemClient.on('error', console.error);

        this.atemClient.connect(atemIP);
    }
}
```

josephdadams:
This is great. I am flexible on what we provide back to the main application - each source type is going to have differing types of busses: pvw, pgm, multiple auxes, perhaps keyers, even other ME's which would be multiple pvw, pgm, etc. So I think if we can provide this back to the main application in an array, where the user can then choose which bus they want to associate with their Device when assigning the address, that would be good. The end goal could be that the user can say "ATEM 'Input 1' on ME 1 PGM maps to TA Device 'Camera 1' PGM". What do you think?

hrueger:
Sounds good. Another question: Are you sure that the ATEM uses port 9910? As far as i know, the ATEM client from the lib just opens a tcp connection some port at the atem's side but does not start a server on the client side. Or am I wrong?

josephdadams:
You're right, so no need to lock down port 9910. That is, until we write ATEM emulation for TA so that it presents itself like an ATEM on the network, enabling support for any existing ATEM tally client. :)

hrueger:
@MatteoGheza @josephdadams @JTF4 I'm stuck at a weird issue for quite some time now. It probably helps if someone else could take a quick look?

I'm currently working on the Tally data flow. The way it is supposed to work is that the source (currently only the OBS source class in the `sources/OBS.ts` file does that) updates busses of addresses using `addBusToAddress()`, `removeBusFromAddress()`, `removeBusFromAllAddresses()` or `setBussesForAddress()` (all defined in the parent class). This calles `this.tally.next(data)` which is an RXJS Observable. The data is an object where the keys are the source addresses and the values are the busses for each address as `string[]`. 

That is then handled [here](https://github.com/josephdadams/TallyArbiter/blob/7c11d7a98e58fa1d23a4486b18ea7c8cbfeb4d2a/src/index.ts#L2039). However, with the OBS Source in particular, I have the strange behaviour that not only `TestImageA` and `TestImageB` are added to the tally data (in the source class) but also `ef8211cc` (which is the `id` of the `device_source` in my `config.json`).
This is not correct, because the source class (OBS) should not know about that `device_source` and no other part of TA should manipulate its tally data. Any idea?

JTF4:
Ill be honest, I really don't know typescript in depth yet, but I'll look at it next time im on my computer.

JTF4:
Ill be honest, I really don't know typescript in depth yet, but I'll look at it next time im on my computer.

hrueger:
Never mind, found it: I was manipulating the wrong object 🤦‍♂️

hrueger:
@josephdadams Question about the bus linking: in the UI we currently set `linkedPreview` or `linkedProgram`. Should we make this dynamic per bus, too? And what does the `linkedBusses` property of a `device` in the main application backend do?

josephdadams:
Linkedbusses is the new replacement. It’s an array of bus id’s and if the bus id is found in the array it’s assumed to be linked. Needs to be implemented in the UI.

hrueger:
@MatteoGheza If we bundle all source files into a single one using webpack, I think we can get rid of [those lines](https://github.com/josephdadams/TallyArbiter/blob/7be49f42c1b0b96f34e07aa59778957a20f42428/src/index.ts#L43-L45). They are just there to load the source class files, because then the decorators are executed and the source types are registered. However, we need that for development.

hrueger:
What are you currently working on? Just that we're not working on the same things. My next points are:

- [ ] Give Tally Data to UI
- [ ] Move Actions into separate classes (with more decorators ;-) )
- [ ] Move remaining Source Types ito their classes

I do probably need help with the last one as I don't completely understand how every source type works.

MatteoGheza:
> @MatteoGheza If we bundle all source files into a single one using webpack, I think we can get rid of [those lines](https://github.com/josephdadams/TallyArbiter/blob/7be49f42c1b0b96f34e07aa59778957a20f42428/src/index.ts#L43-L45). They are just there to load the source class files, because then the decorators are executed and the source types are registered. However, we need that for development.

I'm working on bundling every file in "sources", so you can keep that lines.

MatteoGheza:
> What are you currently working on? Just that we're not working on the same things. My next points are:
> 
> * [ ]  Give Tally Data to UI
> * [ ]  Move Actions into separate classes (with more decorators ;-) )
> * [ ]  Move remaining Source Types ito their classes
> 
> I do probably need help with the last one as I don't completely understand how every source type works.

Currently I'm working on fixing the CI (Webpack).

MatteoGheza:
Importing sources in Webpack doesn't work yet, I'll fix that ASAP.

MatteoGheza:
I'm trying to fix Webpack, but it doesn't work and source classes aren't imported.
@hrueger any suggestions?

hrueger:
Hm, I don't have a lot of experience with webpack, I hardly ever use it on my own projects. What advantages does it have over just using the typescript compiler (`tsc`)? I know about minimization and probably optimization but I think for now we can just compile using tsc.

MatteoGheza:
@hrueger I had not used `tsc` yet. It's awesome!
I replaced `Webpack` with `tsc` since it's really fast and it works well with this code.

hrueger:
@MatteoGheza The models and types that live in the `_models` folder are needed in the backend as well as in the UI. I haven't found a way to share the same file between both, so I created https://github.com/hrueger/redundancyjs (shameless plug 😉). Wehn running `npx redundancyjs` it copies the models from the api to the UI and adds a comment that we should not edit the models in the `UI` folder.

MatteoGheza:
> @MatteoGheza The models and types that live in the `_models` folder are needed in the backend as well as in the UI. I haven't found a way to share the same file between both, so I created https://github.com/hrueger/redundancyjs (shameless plug 😉). Wehn running `npx redundancyjs` it copies the models from the api to the UI and adds a comment that we should not edit the models in the `UI` folder.

What about:

- Adding a symbolic links between  `_models` and  `UI/src/app/_models`. *But* Windows doesn't support git syslink if I remember correctly...
- Adding a README  `UI/src/app/_models`, adding the entire dir (except README) and adding `redundancyjs` to `devDependencies` and to `build` script in `UI/package.json`

hrueger:
I like the second approach. If we add the script to `postinstall`, it will be run after `npm install`.

hrueger:
@josephdadams What are the `source_types_busoptions` for? Currently they are neither used in the UI nor in the main application.

hrueger:
Ah, the when adding a device source to the Carbonite, there's another field in the UI, the `Bus`. I only have an ATEM, would this make sense for the ATEM, too? (Just for me to understand what it does).

hrueger:
The address selection box in the UI (when adding a new device source) is now populated by a new data field called `addresses`. Those are per source, not per source type. They are not hardcoded, but each source class can add addresses using `this.addAddress(label, address)`. 
This allows us to do the following: When adding a new source (for example an ATEM) we can ask the user what model he has (Mini, Mini Extreme, TV Studio, ...) and then add the respective addresses. This will make it more easy for users because then they can just select the address and don't need to type it in.

hrueger:
If I read the `v3.0.0` Project by @MatteoGheza correctly, this PR is targeted for TA 3.0.0. If so, @josephdadams can I remove some legacy code? I was thinking about removing the old separate listener client connection api (`device_listen_m5`, `device_listen_gpo`, `device_listen_relay` and `device_listen_blink`). 
And do we need to keep the REST API?

Don't get me wrong, we can totally keep it, I'm just wondering.

hrueger:
And (sorry for so many comments): What is Smart Tally Status?

hrueger:
In b73d419 I migrated the Blackmagic VideoHub source into its own class. Could someone please check that this actually works as I don't have a VideoHub to test against.

josephdadams:
> @josephdadams What are the `source_types_busoptions` for?  ...  Same for `source_types_busaddresses`

They are only for `Ross Carbonite` models at the moment. It would be fine with me to move these elements directly into the new class for `Ross Carbonite`. It is basically the input for being able to process the extra data that comes in the TSL 3.1 feed from Carbonite models, so that the user can monitor any ME/Aux/etc. rather than just the "on air" setting of the switcher. It makes sense for it to live in the Carbonite class, because it is only related to that source.

josephdadams:
> can I remove some legacy code? I was thinking about removing the old separate listener client connection api (`device_listen_m5`, `device_listen_gpo`, `device_listen_relay` and `device_listen_blink`).
> And do we need to keep the REST API?

Yes, we can remove those old APIs once we have those listener clients updated to work with the new API. It will take some testing because there are some limitations on the client end due to varying implementations of the socket.io protocol. I will do some work on those clients to make this possible. @JTF4 is working on a new m5stick client and once that is done, we can port it for the m5atom.

We will probably want to coordinate a larger release so that people do not try to download 3.0.0 of the server without having the correct version of the listener clients. I don't mind that they will have to upgrade all their clients, I just want to make sure that when we do the release, it's all available at the same time. I may go ahead and implement a version check for each listener client and that way, if anyone uses that code before 3.0.0 is released, it will try to be somewhat backwards compatible, or at the very least, notify the user.

Fine with me to remove the REST API, no one is using it as far as I know, and the last I checked, I am not even sure it is working properly with the new UI. A very initial version of TA relied on this REST API. The UI is so feature rich now that I cannot imagine someone would want to manipulate their TA data through the REST API. Be sure to remove the API data from the documentation as well.

josephdadams:
> And (sorry for so many comments): What is Smart Tally Status?

TA has the `Roland Smart Tally` source type, for getting data from a Roland video switcher that uses their protocol.

TA also emulates the same Roland Smart Tally protocol, on port `4455`, which would enable any compatible Roland Smart Tally software/project to work with TA. See https://github.com/Xylopyrographer/STAC. Typically Smart Tally runs on port `80` but we don't want to deal with the privileged port.

josephdadams:
> I migrated the Blackmagic VideoHub source into its own class. Could someone please check that this actually works as I don't have a VideoHub to test against.

I have a VideoHub, Ross Carbonite (original model, not the Black, Ultra, etc.), and ATEM Production Studio, for testing.

I can get us access to Tricaster if needed.

josephdadams:
@MatteoGheza just fyi a lot of those unused functions that were removed contains the logic for each source type and will need to be brought into the respective classes. :)

MatteoGheza:
@josephdadams I reverted the commit

JTF4:
I think we should add a warning when an outdated listener client tries to connect. Nothing special, just a pop up with information about the new clients.

josephdadams:
We might be able to do something where when a client connects, we start an interval timer, and if the client doesn’t announce its version within the alotted time, we can flag it and put up a caution icon or something next to it in Settings. The reason for this approach is that none of the existing clients currently announce or request any version info, so we have no way of knowing.

hrueger:
There's now not only the `UsesPort(port)` decorator but also the `UsePort(port, sourceId)` function. The former should be used to decorate classes (with the `@` prefix). The latter can be used procedurally in code, for example, when a dynamic port is used. At the end of the lifecycle of that class or whatever it may be, you need to call `FreePort(port, sourceId)` to unregister this port from the `PortsInUse`. The `UsesPort` decorators does not have a free-function because those ports will always be used.

hrueger:
I also migrated the `TSL 3.1 TCP`, `TSL 3.1 UDP`, `TSL 5 TCP` and `TSL 5 UDP` sources. Note that there are some `//ToDo` comments. Whoever knows how the TSL works (with tally1-4) feel free to fix that up and finish the TSL sources.

By the way: if there are any questions about how something now works, why I did it the way it is now or whatever else, feel free to ask :-)

hrueger:
I have migrated all the other sources, too. Let's track their status here:
<table>
<thead>
  <tr>
    <th>Source<br></th>
    <th>Status<br></th>
    <th>Tested by</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td>Analog Way Livecore</td>
    <td>needs testing</td>
    <td></td>
  </tr>
  <tr>
    <td>Blackmagic ATEM</td>
    <td>not implemented completely, will be done by @hrueger soon</td>
    <td></td>
  </tr>
  <tr>
    <td>Blackmagic VideoHub</td>
    <td>needs testing</td>
    <td></td>
  </tr>
  <tr>
    <td>Incoming Webhook</td>
    <td>missing implementation<br></td>
    <td></td>
  </tr>
  <tr>
    <td>Internal Test Mode</td>
    <td>missing implementation<br></td>
    <td></td>
  </tr>
  <tr>
    <td>Newtek Tricaster</td>
    <td>needs testing</td>
    <td></td>
  </tr>
  <tr>
    <td>OBS</td>
    <td>needs more testing</td>
    <td>@hrueger</td>
  </tr>
  <tr>
    <td>OSC</td>
    <td>needs testing</td>
    <td></td>
  </tr>
  <tr>
    <td>Panasonic AVHS410</td>
    <td>needs testing</td>
    <td></td>
  </tr>
  <tr>
    <td>Roland Smart Tally</td>
    <td>needs testing</td>
    <td></td>
  </tr>
  <tr>
    <td>Ross Carbonite<br></td>
    <td>needs testing</td>
    <td></td>
  </tr>
  <tr>
    <td>TSL 3.1 &amp; 5</td>
    <td>not implemented completely, see `//ToDo` comments</td>
    <td></td>
  </tr>
  <tr>
    <td>VMix<br></td>
    <td>needs testing</td>
    <td></td>
  </tr>
</tbody>
</table>

JTF4:
I should be able to test tsl later today.

hrueger:
TSL should work now, too. I filled the missing parts by inspecting the old `index.js` as well as some guessing ;-)

MatteoGheza:
@hrueger OBS wasn't working for me. It threw `TypeError: Cannot read property 'id' of undefined`.
The error is here ( `src\index.ts:1626` ):
Code_fuOk1AXd6h  

Using ~some console.log()~ an advanced debugger and I found the error: `device_sources.find((s) => s.sourceId == source.id && s.address == sourceAddress)` can be undefined.

MatteoGheza:
Anyway I tested OBS with a test collection of scenes.
- `Webcam scene` with `Webcam capture`
- `Canon scene` with `Canon capture`
- `Webcam + Canon` with an example text and two nested scenes (`Webcam scene` and `Canon scene`)

I found two problems: when TallyArbiter starts, the current state of sources isn't captured.
Besides, it's not working for nested scenes. I'm working on it.

hrueger:
@MatteoGheza great, thanks for testing 👍 Yes, the current state is not read when TA is started because we're just listening for the change events.

hrueger:
I added an issues column to the sources list as well as a new table for the actions above.

MatteoGheza:
Done! Support for nested scenes in OBS!
@hrueger can you test it?

hrueger:
All merge conflicts should now be resolved. But please don't push to master until this is merged, because I had to add some commits manually. They created the conflicts and needed to be changed to typescript anyway (dbdea91, 05254d7, 1e88dbc, 802c7b5 and 7cbafd7).

hrueger:
@MatteoGheza the tests fail because this `versions` file in `environments` is not found (it is excluded in the `.gitignore`). Can you explain how that is supposed to work or what it does (or maybe even fix it up)?
I'll test OBS nested scenes tomorrow.

MatteoGheza:
> @MatteoGheza the tests fail because this `versions` file in `environments` is not found (it is excluded in the `.gitignore`). Can you explain how that is supposed to work or what it does (or maybe even fix it up)?
> I'll test OBS nested scenes tomorrow.

`versions.ts` is generated by `git.version.js`

hrueger:
Oh, is that called in a script in a `package.json`? Maybe I missed that while resolving conflicts... sorry. Can you fix that?

MatteoGheza:
There were two `prebuild` scripts in `package.json`.

MatteoGheza:
There's an other problem in CI, I'm working on it.

MatteoGheza:
Now it's working.

JTF4:
There still seems to be a few issues with the TSL implementation, but I'm going to work on that tomorrow.

MatteoGheza:
@jmcc94 I'm sorry, I'm late.
Today I added the requested feature and I added support for VirtualCam and ReplayBuffer.
Besides, now TallyArbiter loads OBS sources status (if they are in preview or in program) at the start.
These contributions are in a bigger pull request, that includes lots of features and improvements.
I'll close this issue only when we release a version with these features.

hrueger:
@MatteoGheza I can confirm that nested scenes are now working. Very cool. I think that for a perfect OBS source only #212 is missing, but in my opinion that's rather low priority.

MatteoGheza:
I'm working on it.

> @MatteoGheza I can confirm that nested scenes are now working. Very cool. I think that for a perfect OBS source only #212 is missing, but in my opinion that's rather low priority.

I'm working on it.

src/index.ts
```diff
@@ -617,173 +445,11 @@ function initialSetup() {
 			}
 		});
 
-		socket.on('device_listen', function(deviceId: string, listenerType: string) { // emitted by a socket (tally page) that has selected a Device to listen for state information
```
MatteoGheza: This socket event is used in "Tally" page that now is not working.

MatteoGheza (changes requested):

src/index.ts
hrueger: I'll update the Web Tally. It should also use the new connection api with the `listenerclient_connect` event.

hrueger:
And we have the index.ts down to less than 2700 lines 😀

MatteoGheza:
Can you fix Tally page?

hrueger:
Should work now. Can you try it?

MatteoGheza:
Tally page is working.

hrueger:
@josephdadams I added a new event for the listener clients, called `deviceTallyData`. It is send when the tally data for that device changes. The only data that comes with it is an array of bus ids (strings).
The Web Tally is already updated accordingly.

@MatteoGheza Sorry, can you test that again? I just changed some logic, because we were sending all tally information to the clients previously and they had to pick their device.

MatteoGheza:
#212 done


https://user-images.githubusercontent.com/27729549/132654139-298e6d3e-dfb7-4bab-b358-cf46b45a7bb5.mp4



https://user-images.githubusercontent.com/27729549/132654086-f138b64a-bf4d-4fce-9855-b849c7d80497.mp4


Is this ok?

hrueger:
@josephdadams What do you think about moving the listener clients into this repository? That would make it a monorepo. We could control releases more easily by also releasing new listener client firmware / software. I don't even know what listener clients are all out there.

hrueger:
>Is this ok?

Sure, it looks great 👍👍

MatteoGheza:
@hrueger renaming sources in OBS doesn't work. It doesn't crash TallyArbiter, and it updates addresses with the new address, but UI has the old address so it doesn't work.
Can you take a look at it?

hrueger:
Yes, that not inplemented currently. I'll fix that as soon as my laptop boots again 😐

MatteoGheza:
Currently if source is unmuted it is assigned to program, and if not its bus list is empty. Is this ok?
Should I had to assign muted sources to "preview"?

hrueger:
I think its perfekt like this (program or nothing) 👍

josephdadams:
> @josephdadams What do you think about moving the listener clients into this repository?

There are several.

- M5 Stick C/ C+
- M5 Atom
- Blink(1)
- Relay
- GPO
- Pimoroni Blinkt
- TT_GO
- ESP32/NeoPixel
- ESP8266/NeoPixel

And with options like Vmix emulation, Roland emulation, etc., there are many more clients that TA can natively work with.

I am not opposed to discussing moving them into this repo but not currently convinced. Can we bring that into a new issue to discuss?

josephdadams:
> @josephdadams I added a new event for the listener clients, called `deviceTallyData`. It is send when the tally data for that device changes. The only data that comes with it is an array of bus ids (strings).

Just to help me understand, what is the advantage of doing it this way?

hrueger:
Sorry, I must have deleted the part with the listener updates while refactoring and therefore created a new event. 
Now I found that part in the old `index.js` again. I'll change it. The old event was called `device_states`, right?
Is it ok if we send an array of active bus ids with that event?

josephdadams:
`device_states` is essentially an array of:

- deviceid
- busid
- array of sources currently active in that bus

So there will be multiple entries of the same deviceid, but the unique key is deviceid + busid.

This means that the `device_states` array contains at all times all active sources on all active busses. So if we can filter that to just the deviceid the listener client needs, it would allow, at the listener client level, to implement a solution that even shows the user "the device is in program on all of these sources" for example.

So while it's more data, it's not any harder to process, and it's more verbose, or at least opportunity to be more verbose at the client level. This is how all of the listener clients currently process - they receive a filtered `device_states` (only for their deviceid). They also receive the `bus_options` array, so they know everything about that bus, like color, priority, etc.

hrueger:
Ah, that makes sense. I'll implement that again 👍

josephdadams:
The `busoptions` branch on the blink(1) listener has the latest bus options, color, and priority implemented if you want to look at that for the web tally listener.

hrueger:
> `device_states` is essentially an array of:
> 
> - deviceid
> - busid
> - array of sources currently active in that bus

Does that mean that if the device leaves preview, and is now in program but not in preview anymore, I send that object for the program bus with the active source as well as the preview bus with an empty source array?

And what properties do we send in the source array? Label, address and ID?

josephdadams:
Any time the device states change for any reason, resend the entire device states. Or a filtered version for only that device id if you wish. 

The source property array only contains sourceids and we can match that to the sources array if needed. I’m not sure sources are currently being sent, but if a listener client asks for it, TA can send it back.

hrueger:
Now the `device_states` event is used again.

JTF4:
I noticed that you had to reload the page when you delete a source in order for the port to be reported as open. I added a socket update for `portsinuse` on source deletion

I also fixed the `freeport` on the exit functions in the TSL source

Also for removing a used port, I had to remove the `PortsInUse.next` in order for it to keep working`.

hrueger:
May I recommend using `npm@7` with `node@14` or later? `npm@7` uses the lockfile version 2 while `npm@6` uses the old lockfile format. In a lot of commits (for example bc2834d) the `package-lock.json` files of the application and the UI are updated, because some of us use the `npm@6` and some `npm@7`.

hrueger:
@JTF4 Thanks for fixing the `PortsInUse`. I slightly modified that again. The reason for it not working was that the `splice` method apparently returns the deleted elements and modifies the original array. So now that part looks like this:
```TypeScript
PortsInUse.value.splice(PortsInUse.value.findIndex((p) => p.port == port && p.sourceId == sourceId), 1)
PortsInUse.next(PortsInUse.value);
```

This removed the need for calling `UpdateSockets('PortsInUse');` as this is done automatically for every change of `PortsInUse` [here](https://github.com/josephdadams/TallyArbiter/blob/b51d774f778e78b483e08384c52e92fed396c422/src/index.ts#L123-L125).

I hope you're ok with that.

hrueger:
@josephdadams How does [this function](https://github.com/josephdadams/TallyArbiter/blob/a318a0995e444fbcc09349a8180bbe0a47201d3a/index.js#L6386-L6450) work? How does a vMix Listener Client specify what Device it wants to listen to?

josephdadams:
@hrueger vmix tally only supports preview/program, so it sends all tally data to all connected clients like you saw, `02102` where `0` is off, `1` is pgm, and `2` is pvw. The order is index order of the Devices array, that small number we display in the UI next to the device name which is `index+1` for display purposes. Each VMix tally client is individually configured to then listen to a specific VMix input. At the server level we have no real way of knowing which tally they are monitoring since it is done internally on the client end.

josephdadams:
> May I recommend using `npm@7` with `node@14` or later? `npm@7` uses the lockfile version 2 while `npm@6` uses the old lockfile format. In a lot of commits (for example [bc2834d](https://github.com/josephdadams/TallyArbiter/commit/bc2834d7d8e5ec17dc0bce00816c1b1bb2892ecf)) the `package-lock.json` files of the application and the UI are updated, because some of us use the `npm@6` and some `npm@7`.

Let's be sure to mention this in the dev requirements of the readme for anyone who wants to contribute, that we require `npm@7` and `node@14`.

hrueger:
I added another table to [this comment](https://github.com/josephdadams/TallyArbiter/pull/204#issuecomment-915174735) with more stuff to test.

I think we're done with refactoring, In my optionion we're ok with the `index.ts` file being at 2280 lines now.

josephdadams:
@hrueger I'll do my best to slowly get it back up to 6000 lines again 😆

hrueger:
`PortsInUse` is an RXJS' `BehaviourSubject` which is basically an [Observable](https://rxjs.dev/guide/observable). The `next(v)` method sets a new value, and everyone who subscribed to changes with the `subscribe` method will be notified (its callback is run). 

We're also using those BehaviourSubjects at other places, for example to provide the connection status of sources.

Does that make sense to you?

JTF4:
Also, what would we think about moving all of the source types and actions into their own folders? That would easily allow us to submodulize them into their own github repos and would also make it a little more user friendly to contribute to as there would be a local _models folder for each thing.

hrueger:
Hm. I am (personally) a fan of monorepos, because everything is in one place. What advantages do you see of multiple repos and what do you mean by _as there would be a local _models folder for each thing_? 
I also don't understand how you can contribute using just the (for example) ATEM souce type repo, because you can't test it there, can you?

JTF4:
On second thought, I agree with you on monorepos. I was thinking of companion, but we wont have anywhere close to the amount of integrations that they have.

The only thing creating repos would do is give you a better way of developing new integrations where any local types you need to create are contained within the modules. Eg. you need a custom array to process in the incoming information from a device, but instead of adding it to the global `_models` folder, you could add them to a `_models` folder contained within your integration's file.

josephdadams:
@hrueger I haven't had a chance to look so I apologize if it's already done. But is there a way, in the UI, when adding Device Sources (assigning the incoming Source tally address to a Device) to also choose the bus it's coming from in the Source, and map it to the Bus in TA? I think this is the next big area of logic that we will have to sort out if we want to truly support multiple busses. Some users will want to have their Camera 1 listed in PGM when it's in PGM on their ATEM, but perhaps listed in Aux 1 when it's in PGM on their OBS.

hrueger:
You're right, that's not possible yet. Unfortunately, school has started for me again yesterday, so I don't have a lot of time for open-source work... Maybe someone else can implement this?

josephdadams:
I understand. We may just have to scale that feature back for now. It won't change the implementations of the bus options with priority levels, it just means there won't be any other busses than pvw/pgm. I'm ok with that.

hrueger:
Great. For now, sources can set addresses to aux busses by themselves, too.

JTF4:
Before I would release a beta I would fix internal test mode, because at the moment, it does not work.

hrueger:
Unfortunately, I don't have time to work on that...

JTF4:
I think we can just re-implement the same logic that ran in the original version of TA. We can implement all of the speed controls and whatnot later.

mattv8:
Finally looking into updating my C# client to work with the new custom listener api. Can anyone point me to a commit or lines of code that implement the new api? Are you doing away with the `socket.on('device_listen', function(deviceId, listenerType)` paradigm entirely?

MatteoGheza:
@mattv8 I think that the new API is using `listenerclient_connect` socket event.
See lines 245-299 in `src/index.ts`.
Here that's an excerpt:
```js
socket.on('listenerclient_connect', function(obj: ListenerClientConnect) {
    /*
    This is the new listener client API, all clients should connect and send a JSON object with these properties:
    deviceId
    listenerType (string to be displayed)
    canBeReassigned (bool)
    canBeFlashed (bool)
    supportsChat (bool)
    */

    let deviceId = obj.deviceId;
    let device = GetDeviceByDeviceId(deviceId);
    let oldDeviceId = null;

    if ((deviceId === 'null') || (device.id === 'unassigned')) {
        if (devices.length > 0) {
            oldDeviceId = deviceId;
            deviceId = devices[0].id;
            socket.emit('error', 'Invalid Device Id specified. Reassigning to the first Device on the server.');
        }
        else {
            //send error state
            socket.emit('error', 'No Devices are configured in Tally Arbiter.');
        }
    }

    let listenerType = obj.listenerType ? obj.listenerType : 'client';
    let canBeReassigned = obj.canBeReassigned ? obj.canBeReassigned : false;
    let canBeFlashed = obj.canBeFlashed ? obj.canBeFlashed : false;
    let supportsChat = obj.supportsChat ? obj.supportsChat : false;

    socket.join('device-' + deviceId);
    let deviceName = GetDeviceByDeviceId(deviceId).name;

    if (supportsChat) {
        socket.join('messaging');
    }
    else {
        socket.leave('messaging');
    }

    socket.emit('bus_options', currentConfig.bus_options);
    socket.emit('devices', devices);
    socket.emit('device_states', getDeviceStates(deviceId));

    if (oldDeviceId !== null) {
        //sends a reassign command to officially reassign the listener client to the new device ID since the first one was invalid
        ReassignListenerClient(clientId, oldDeviceId, deviceId);
    }
});
```

MatteoGheza:
BTW @JTF4 can you test my new internal test mode implementation?
It doesn't currently use "Test mode button", since is still WiP, but it's a standalone source, that you should add to TallyArbiter like any other source.
What do you think? Should I integrate this with "Turn on/off test mode" button (for example, showing and hiding Internal Test Mode in sources list)?

mattv8:
@MatteoGheza beautiful, exactly what I needed to know. Thanks for the info!

JTF4:
@MatteoGheza It works great. Are we going to make it automatically as itself as a source to all of the devices?

In terms of the button, I would make the button add the source and take it away, but also put all of the settings in the testing page. I also think that we should add an event to the button that makes the state reflect across all settings pages that are open on devices.

hrueger:
I added some remaining things:
- Fields of Source Types can now be marked as optional. This disables the check if all fields are filled out in the UI. Required (non-optional) fields are marked with a red asterisk *.
- I moved the _Test Mode_ Button to it's testing tab and also show the Devices table again there. 
- When enabling the Test Mode, Device Sources are automatically added to all devices, removing the need for the user to configure this by hand.

I think we can merge this and then release a beta?

josephdadams:
Great! Yes I think we can release a beta and then let issues bubble up.

josephdadams:
We will need to push updates to all the listener clients to use the new API

MatteoGheza:
I copy-pasted some comments in #224 so we can track our progress in v3.0.0

hrueger:
Great. If you want, you can do the merge and the beta release, because I won't have time for that today.

JTF4:
@MatteoGheza test mode works great. The only thing is that the interval does not change the speed of the test that's running.

JTF4:
I'm working on the interval now

JTF4:
Internal test mode is up and running

JTF4:
When testing the OSC action I'm getting this stack trace:

`TypeError: args_1[i].replace is not a function
    at OSC.run (/Users/davidstevens/Documents/GitHub/TallyArbiter/src/actions/OSC.ts:48:40)
    at RunAction (/Users/davidstevens/Documents/GitHub/TallyArbiter/src/index.ts:1183:13)
    at _loop_1 (/Users/davidstevens/Documents/GitHub/TallyArbiter/src/index.ts:996:6)
    at UpdateDeviceState (/Users/davidstevens/Documents/GitHub/TallyArbiter/src/index.ts:931:13)
    at processSourceTallyData (/Users/davidstevens/Documents/GitHub/TallyArbiter/src/index.ts:1163:3)
    at /Users/davidstevens/Documents/GitHub/TallyArbiter/src/index.ts:1132:3
    at Object.next (/Users/davidstevens/Documents/GitHub/TallyArbiter/node_modules/rxjs/src/internal/Subscriber.ts:194:14)
    at SafeSubscriber.Subscriber._next (/Users/davidstevens/Documents/GitHub/TallyArbiter/node_modules/rxjs/src/internal/Subscriber.ts:119:22)
    at SafeSubscriber.Subscriber.next (/Users/davidstevens/Documents/GitHub/TallyArbiter/node_modules/rxjs/src/internal/Subscriber.ts:75:12)
    at /Users/davidstevens/Documents/GitHub/TallyArbiter/node_modules/rxjs/src/internal/Subject.ts:63:20`

hrueger:
@MatteoGheza what's your current status? Do you have an ETA for merging this? I'd then create a few PRs for Hacktoberfest, too ;-) but I'd like to have this merged before.

josephdadams:
It’s fine with me to merge and then we can just fix issues as they come up. We have stable releases people can use in the mean time.

hrueger:
OK, then I'm going to merge them tomorrow.

PR #241, opened 2021-10-17
Show outdated browser info

### LINKED ISSUE #116: Show "Your browser is not supported" message if browser is too old
> Is there a way to get the library to at least leave "Sorry; your browser is too old; please upgrade" in the returned window?

We should use https://github.com/browserslist/browserslist-useragent-regexp and some plain js in the `index.html` file to output a warning if the browser is unsupported. Currently users just get a blank page.

_Originally requested by @Baylink-gh in https://github.com/josephdadams/TallyArbiter/issues/115#issuecomment-883546988_

hrueger:
fixes #116

I found this really handy site called https://browser-update.org. 

It works perfectly for me in Internet Explorer.

Do you mind adding this (kind of) third party dependency?

In the future we could implement this ourselfes for offline functionality, but I think for now it's just fine.

### DIFF
--- a/UI/src/index.html
+++ b/UI/src/index.html
@@ -21,5 +21,15 @@
 <body>
   <app-root></app-root>
   <noscript>Please enable JavaScript to continue using this application.</noscript>
+  <script> 
+    var $buoop = {required:{e:89,f:88,o:78,s:13,c:90},insecure:true,api:2021.10 }; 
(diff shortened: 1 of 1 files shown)


josephdadams:
If the device is so old and they don’t have internet they will probably figure that out. So I’m fine with it.

PR #244, opened 2021-10-18
Add sentry.io error logging

JTF4:
Sentry.io captures and logs errors in the cloud so that we know what the biggest issues are in TA


JTF4:
@josephdadams thoughts on this PR?

josephdadams:
I’m good with it as long as it’s opt in not opt out. Turned off by default.

JTF4:
Users are now opted out be default

JTF4:
@MatteoGheza it works great!

PR #260, opened 2021-10-29
Added CODE_OF_CONDUCT.md

Jassi10000:
Added CODE_OF_CONDUCT.md to the repo

@MatteoGheza @JTF4 
Sir can you please review and merge this


Jassi10000:
@hrueger  and @josephdadams Please review the latest changes

hrueger:
@Jassi10000 thanks for the update. However, as the files are now deleted, when we merge the PR, we lose those files in the main branch, too...
So it would be great if you could just copy those files from the main branch here and commit that. Then, this PR should only show "1 file changed"

hrueger:
@MatteoGheza I'd prefer the "squash and merge" strategy for this PR, because then all the commit are squashed into a single one and we don't have the removal and adding back of the `package.json` files in our history.

Jassi10000:
Should I close this and make another PR ? @hruger

hrueger:
If you want, you can, but it should also work if you just add back the files.

If we merge using the squash method, only a single commit will appear in the history.

JTF4:
Closing due to #261

PR #292, opened 2021-12-01
Update tallyarbiter-m5atom.ino

wrldtvlr:
Correct color order for M5.dis.drawpix from rgb to grb. Also add code lines to define WiFi SSID and password. Deleted unused brightness line.


MatteoGheza:
@wrldtvlr why did you add WiFi name and PWD in the code?
They aren't used in the codebase and they are useless since now we're using WiFiManager for configuring WiFi credentials.

wrldtvlr:
I guess I'm not familiar with WiFiManager. These were in the v2 client. If they aren't present, how does one specify the SSID and password?

wrldtvlr:
Ahh... Just found a WiFIManager tutorial. I see - if it doesn't connect the client goes into station mode for configuration. I'll pull that code out.

Issue #49, opened 2021-01-29
No recovery after timeout towards vMix


Jocke4f:
### Source code
Latest as of 2021-01-29
#### Environment
Computer: Mac mini (mid 2012)
MacOS: 10.15.7
### Scenario
One vMix source configured. The vMix computer is offline. TallyArbiter setup a TCP port connection towards 8099, but timeouts after 45 seconds with an exception.

    node:events:356
          throw er; // Unhandled 'error' event
          ^
    
    Error: connect ETIMEDOUT 192.168.0.89:8099
        at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1139:16)
    Emitted 'error' event on Socket instance at:
        at emitErrorNT (node:internal/streams/destroy:188:8)
        at emitErrorCloseNT (node:internal/streams/destroy:153:3)
        at processTicksAndRejections (node:internal/process/task_queues:81:21) {
      errno: -60,
      code: 'ETIMEDOUT',
      syscall: 'connect',
      address: '192.168.0.89',
      port: 8099
    }

Due to this must TallyArbiter be launched **after** vMix has been started on the vMix computer. Expectation is that TallyArbiter handles the timout and tries to reconnect until successful.

Jocke4f:
The behaviour, unhandled exception that terminates TallyArbiter Server, is the same if the computer is online but vMix is not running or if vMix is running but is closed.

My expectation is that TallyArbiter should handle such a scenario but I don't know if that is the intention or not.

#### Launching TallyArbiter when vMix is not running on computer
    % node index.js
    [Sat Jan 30 2021 18:12:39 GMT+0100 ()]     Loading the stored Tally Arbiter configuration file.
    [Sat Jan 30 2021 18:12:39 GMT+0100 ()]     Tally Arbiter Sources loaded.
    [Sat Jan 30 2021 18:12:39 GMT+0100 ()]     1 Sources configured.
    [Sat Jan 30 2021 18:12:39 GMT+0100 ()]     Tally Arbiter Devices loaded.
    [Sat Jan 30 2021 18:12:39 GMT+0100 ()]     5 Devices configured.
    [Sat Jan 30 2021 18:12:39 GMT+0100 ()]     Tally Arbiter Device Sources loaded.
    [Sat Jan 30 2021 18:12:39 GMT+0100 ()]     Tally Arbiter Device Actions loaded.
[...]

PR #317, opened 2021-12-25
Configurable reconnect handling

Jocke4f:
The reconnect handling for device sources has been extended with the options:
- reconnect_intervall (number)
- unlimited_reconnects (boolean)

**Reconnect_intervall** is only used if larger than the constant RECONNECT_INTERVAL. 
**Unlimited_reconnects** is only used if set to true. If false or not set the default MAX_FAILED_RECONNECTS is used.

NB: Additional console printouts for easier troubleshooting was included.

### DIFF
--- a/src/_helpers/configSchema.ts
+++ b/src/_helpers/configSchema.ts
@@ -137,9 +137,17 @@ export default {
                     rename: {
                         description: "Enables renaming",
                         type: "boolean"
+                    },
+                    reconnect_intervall: {
+                        description: "Configure reconnect intervall",
+                        type: "number"
+                    },
+                    unlimited_reconnects: {
+                        description: "Enables unlimited reconnect attemps",
+                        type: "boolean"
                     }
                 },
-                required: ["deviceId", "address", "id", "sourceId", "bus", "rename"]
+                required: ["deviceId", "address", "id", "sourceId", "bus", "rename", "reconnect_intervall", "unlimted_reconnects"]
             },
             uniqueItems: true
         },
--- a/src/_helpers/configSchema.ts
+++ b/src/_helpers/configSchema.ts
@@ -138,8 +138,8 @@ export default {
                         description: "Enables renaming",
                         type: "boolean"
                     },
-                    reconnect_intervall: {
-                        description: "Configure reconnect intervall",
+                    reconnect_interval: {
+                        description: "Configure reconnect interval",
                         type: "number"
                     },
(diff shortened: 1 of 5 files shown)


josephdadams:
@Jocke4f thanks! We will give this further review. One thing - can you change your console logs to use our logger function?

Jocke4f:
**Console.log** replaced with **logger** in class TallyInput constructor upon request by @josephdadams.

Jocke4f:
FYI: This PR is related to improve the reconnect functionality that was raised in comments to issue 49 after that the unhandled error events no longer caused the server to crash.

https://github.com/josephdadams/TallyArbiter/issues/49

hrueger (commented):
What if we change `unlimited_reconnects` to `max_reconnects` (or something similar) ? It should be -1 for infinity. Then, on the UI side, we could do a checkbox "unlimited reconnects" which sets the `max_reconnects` to -1. If unchecked, the user could enter a number.
This would give even more flexibility to users.

I'm thinking of situations where all your software is for example in autostart and it takes a few minutes (more than `MAX_FAILED_RECONNECTS`) but if it doesn't connect within a few more minutes (your custom reconnect max), you don't want it to try any longer.

What do you think about that?

Jocke4f:
For us would it work for our scenario since we will still be able to increase the timeout between reconnect attempts and have the server running 24/7. It would however be more flexible and easier to add in to the UI.

I you're all in favour of such a solution can I come back with a new commit with that alternative.

Jocke4f:
The code has now been reworked with the following options:

- reconnect_intervall (number)
- max_reconnects (number)

`reconnect_intervall` is only used if larger than the constant RECONNECT_INTERVAL.
`max_reconnects` is used if set, If not set is the constant MAX_FAILED_RECONNECTS used. -1 is the special case inifinte reconnects.

MatteoGheza (approved):

MatteoGheza:
@hrueger what do you think? Can we merge this?

josephdadams:
It's completely semantics but I would prefer `reconnect_interval` over `reconnect_intervall`, if possible

Jocke4f:
No problem. Renamed `reconnect_interval` to `reconnect_interval` in latest commit.

MatteoGheza (approved):

hrueger (approved):
lgtm. We can add the checkbox in the UI I talked about later on.

PR #321, opened 2021-12-29
Fixed an issue with save config function

JTF4:
When the save config function was called when `useMDNS` was fase, it would reset the `useMDNS` value to true because of the `!server_config.useMDNS` call.


MatteoGheza:
Thanks! I forgot about this.

PR #332, opened 2022-01-01
Support source visibility and OBS Websocket 5

MatteoGheza:


hrueger:
Really nice with the different versions!
Although it's probably still wip, I have a comment: can we check whether studio mode is currently disabled and only enable in that case?
Otherwise, if they change the error message, it counts as an error ;-)

MatteoGheza:
Waiting for "transitions events" and "sources info in `GetScenesList`".

MatteoGheza:
I'll work on this these days.
It's important since now every new OBS install has Websocket 5 preinstalled.
The only problem is the port number, the same used by TA (4455)...

MatteoGheza:
I need to close this PR since refactoring everything would be a difficult work.
I'll use this code for the new implementation, and open a new PR in a few hours/days.

PR #333, opened 2022-01-01
Stability and color fixes M5Atom client

cyberblaststudios:
- Added reconnect logic that will automatically try to re-connect the M5Atom listener to the server if the socket experiences a disconnect event. This is ideal for people that leave the listeners powered on 24/7, so that they don't have to cycle power on M5Atom devices that may be hard to reach if the server ever goes down.
- Added the ability to use a static IP address, mask, and gateway in the code if the user wishes to do so.
- Fixed color formatting issue for the M5Atom devices. M5Atom deals with color in GRB format, instead of RGB, and the server will send RGB hex values to the clients, so code has to compensate for this on the M5Atom.

PR #345, opened 2022-01-11
run workflows on `pull_request` event

hrueger:
This is to make sure that new PRs don't break the CI once they are merged (at least we'll see that jobs fail before merging).
That is probably a good idea also to ensure that workflows don't intruduce build problems.

Created because of #333 which looked ok with the workflows that did run, but once merged it broke ;-)

### DIFF
--- a/.github/workflows/build-listener-clients.yaml
+++ b/.github/workflows/build-listener-clients.yaml
@@ -1,6 +1,6 @@
 name: Build Listener Clients
 
-on: [push, workflow_dispatch]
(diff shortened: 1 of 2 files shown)


josephdadams:
Did you need any feedback on this before merging? I don’t know much about it. So if you feel good, go for it

MatteoGheza (approved):

hrueger:
Thanks @MatteoGheza 
I somehow forgot to merge this ;-)

PR #381, opened 2022-02-20
Remove references to undeclared static IP variables from M5Atom listener

kmyhre:
Static IP functionality for the M5 Atom was partly removed (commented) from PR #333, but the subsequent use of the static IP variables remained, preventing the listener from compiling for the last month. This PR comments out those references as well.

I am not well versed in C++ or Git (at all), so feel free to point out better ways to fix this issue.


JTF4:
@kmyhre I made a PR on your repo that uses a if statement to enable and disable static IPs, if you want to merge that in, I can merge this PR into the main repo

kmyhre:
Thanks @JTF4 – that seems like a much cleaner way of doing it. Merged in my repo!

Issue #305, opened 2021-12-16
[Bug]: New Listener Client API not properly receiving JSON object from client... I think...


mattv8:
### What happened?

In updating my listener client to work with the new Listener Client API, I'm having trouble getting TA to receive my JSON object. I am able to establish the connection just fine, but TA doesn't pull in any of the JSON variables and uses defaults.

From my C# client, the SocketIO emit statement is pretty straightforward (paraphrasing my code for clarity):
```
foreach (Devices device in deviceList) {
    socket.Emit("listenerclient_connect", deviceObj);
}
```
Where `deviceObj` are well-formed JSON strings, and look like the following (using a deviceList containing "Cam 1" and "Cam 2" deviceId's as an example):
```
{"deviceId":"e8a2bd0b","listenerType":"OpenTally Cam 1","canBeReassigned":true,"canBeFlashed":true,"supportsChat":false}
{"deviceId":"9ff13e3d","listenerType":"OpenTally Cam 2","canBeReassigned":true,"canBeFlashed":true,"supportsChat":false}
```

However, TA establishes the connections with default listenerclient_connect variables:
![image](https://user-images.githubusercontent.com/9312603/146434938-1783433c-fd5a-46be-b924-4c2065924e71.png)

I have also tried slightly different client syntax like the following with the same issue:
```
foreach (Devices device in deviceList) {
    client.listenerType = "OpenTally " + device.name;
    client.canBeReassigned = true;
    client.canBeFlashed = true;
    client.supportsChat = false;
    socket.Emit("listenerclient_connect", device.id, listenerType, canBeReassigned, canBeFlashed, supportsChat);
}
```

Any tips to help me debug?

### Version

3.0.1

### Distribution

Desktop

### OS

Windows

### TallyArbiter configuration

```javascript
{
  "security": {
    "username_producer": "producer",
    "password_producer": "12345",
    "username_settings": "admin",
    "password_settings": "12345"
  },
  "cloud_destinations": [],
  "cloud_keys": [],
  "device_actions": [],
  "device_sources": [],
  "devices": [
    {
      "name": "Cam 1",
      "description": "Cam 1 Tally",
      "linkedBusses": [
        "e393251c",
        "334e4eda"
      ],
      "enabled": true,
      "id": "e8a2bd0b"
    },
    {
      "name": "Cam 2",
      "description": "Cam 2 Tally",
      "linkedBusses": [
        "e393251c",
        "334e4eda"
      ],
      "enabled": true,
[...]

PR #393, opened 2022-03-02
BUGFIX: Custom listenerClient API

mattv8:
This PR seeks to fix a bug I found with the Custom listenerClient API. Addresses issue #305

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -332,9 +332,10 @@ function initialSetup() {
 			supportsChat (bool)
 			*/
 
-			logger(`Received JSON object: ${obj}`); //To make it easier for devs to verify the JSON objects they're sending
-
-			obj = JSON.parse(String(obj)); //Re-parse JSON
+			if(typeof obj !== 'object' && obj !== null) {
+				logger(`Received JSON object: ${obj}`, 'info-quiet'); //Log the raw JSON to console
+				obj = JSON.parse(String(obj)); //Re-parse JSON
+		   	}
 
 			let deviceId = obj.deviceId;
 			let device = GetDeviceByDeviceId(deviceId);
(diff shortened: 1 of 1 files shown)


JTF4:
I think this looks good. However, I'm going to do some digging to see why the JSON object is not parsed at the beginning to make sure we don't have a bigger issue.

src/index.ts
```diff
@@ -331,10 +331,11 @@ function initialSetup() {
 			canBeFlashed (bool)
 			supportsChat (bool)
 			*/
-			obj = JSON.parse(String(obj)); //Re-parse JSON
 
 			logger(`Received JSON object: ${obj}`); //To make it easier for devs to verify the JSON objects they're sending
 
+			obj = JSON.parse(String(obj)); //Re-parse JSON
```
JTF4, line 337: I would change this to 

`if(typeof obj !== 'object' && obj !== null) {
     logger(`Received JSON object: ${obj}`, 'info-quiet');
     obj = JSON.parse(String(obj)); //Re-parse JSON
}`

That way it does not parse an already parsed JSON object

JTF4 (changes requested):

JTF4:
@mattv8 if you can get this finished up today, I can get it included in the new beta release later today

mattv8:
Good to go @JTF4. I wanted to take the time to verify with my software and listener clients first. Great suggestion to add a check before reparsing.

Be aware, this may also be an issue for the `listener_reassign_object` function as well, however we have the `listener_reassign` which doesn't take a JSON object-- that's been working well for my purposes.

JTF4 (approved):
Looks great! I'll test the other two functions you're referring to and see what they do

PR #424, opened 2022-04-12
RGB to GRB bug correction

monkzzz:
The code wasn't converting the RGB to GRB.
It maintained the RGB in original code and displayed the colors Green and Red swapped.

With this correction the colors are now in GRB form and displayed correctly.


josephdadams:
solves #409

josephdadams:
solves #409

PR #437, opened 2022-05-05
MDNS Improvements

### LINKED ISSUE #421: [Bug]: BMD Devices being falsely detected as ATEMs
### What happened?

We have several BMD devices detected via mdns that are not actually ATEMs:
Screen Shot 2022-04-07 at 10 44 16 PM

### Version

3.0.1

### Distribution

CLI

### OS

Linux

Chrome

### If applicable, What Listener Clients are You Using?

_No response_

### TallyArbiter configuration

_No response_

### Relevant log output

_No response_

### Error stacktrace (if applicable)

_No response_

hrueger:
- switches from `bonjour` to `bonjour-service` (fixes #421)
- fixes a crash when connecting to an ATEM (this PR is probably not the right place, but I've already committed it ;-) )
- shows the service name in the discovery table (for example to differenciate between ATEM models)

actually, we still need [this](https://github.com/josephdadams/TallyArbiter/blob/7fc90f7f5f8a5725ba5144e11cd3e10f78c80fa1/src/sources/BlackmagicATEM.ts#L17-L18), but once https://github.com/onlxltd/bonjour-service/issues/16 is fixed, we can remove that.


MatteoGheza (approved):
Good job! It should work well now.

PR #442, opened 2022-05-11
Fix #420: vMix Emulator is not started

### LINKED ISSUE #420: [Bug]: vmix emulation not working
### What happened?

No sign of vmix emulation listener port opened. After downgrading to v2 stable, clearing and recreating the config vmix emulation works fine (and of course the 8099 port is visible in netstat's output). 

### Version

3.0.6

### Distribution

Desktop

### OS

Windows

_No response_

### If applicable, What Listener Clients are You Using?

_No response_

### TallyArbiter configuration

_No response_

### Relevant log output

_No response_

### Error stacktrace (if applicable)

_No response_

hrueger:
I don't have a vMix setup here to test this, but I found that we're not even starting the vMix Server ;-)

Could someone test this?

### DIFF
--- a/src/_modules/TSL.ts
+++ b/src/_modules/TSL.ts
@@ -11,8 +11,8 @@ import { Device } from "../_models/Device";
 export class TSLListenerProvider extends ListenerProvider {
     public tsl_clients: TSLClient[] = [];
 
-    constructor() {
-        super();
+    public start() {
+        logger('Starting TSL Listener Service.', 'info-quiet');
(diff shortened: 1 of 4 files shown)


MatteoGheza:
I don't have a vMix setup, but this seems fine.

MatteoGheza (approved):

hrueger:
Do you think we should publish a new release for people to test this as well as the mdns / bonjour fix and the obs improvements?

MatteoGheza:
Yes, I think we should do that

PR #464, opened 2022-07-11
Fix Docker Node Bcrypt error

MatteoGheza:
Closes #463


MatteoGheza:
This can be improved, but it's fixed now.
I removed `Dockerfile.standalone` since it needed a complete rewrite (using different build stages, optimizing build context, ...)

PR #465, opened 2022-07-13
Fix off by one error and unhandled undefined in relay-listener

goseid:
While trying to get relays configured properly and working on a fresh system I came across a state where ProcessTallyData was calling getRelayGroupById with a deviceId that didn't match any of the relay groups. The for loop in getRelayGroupById went through 5 of my 4 relay groups which resulted in comparing the deviceId property on an undefined item from the array and errored out.

After fixing the off by one, the getRelayGroupById function was returning undefined due to the deviceId not matching any groups and ProcessTallyData errored out when trying to compare property deviceId on undefined.


josephdadams:
Good catch! Thank you!

PR #483, opened 2022-09-25
Update m5stickc-listener.ino

commputethis:
Fix internal LED (LOW/HIGH were reversed)

### DIFF
--- a/listener_clients/m5stickc-listener/m5stickc-listener.ino
+++ b/listener_clients/m5stickc-listener/m5stickc-listener.ino
@@ -164,11 +164,11 @@ void setup() {
   #if TALLY_EXTRA_OUTPUT
   // Enable interal led for program trigger
   pinMode(led_program, OUTPUT);
-  digitalWrite(led_program, HIGH);
+  digitalWrite(led_program, LOW);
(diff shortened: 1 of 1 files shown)


listener_clients/m5stickc-listener/m5stickc-listener.ino
```diff
@@ -616,24 +616,23 @@ void evaluateMode() {
...
+    } else if (actualType == "\"aux\"") {
+      digitalWrite (led_program, HIGH);
+      digitalWrite (led_preview, HIGH);
       digitalWrite (led_aux, LOW);
+    } else {
+      digitalWrite (led_program, HIGH);
+      digitalWrite (led_preview, HIGH);
+      digitalWrite (led_aux, HIGH);
     }
-    #endif
```
mrsydster, line 636: This #endif should stay

mrsydster (commented):
small fix of if statement

listener_clients/m5stickc-listener/m5stickc-listener.ino
JTF4: @mrsydster is correct. The #endif should stay.

PR #494, opened 2022-10-17
Fix and refactor OBS Source and add support for OBS 28 (websocket protocol version 5)

MatteoGheza:
Currently, I have fixed tons of bug with the OBS source for OBS<28, and now it should work as requested.
I'm working on supporting the new protocol 5 since now it's a stable version.
Maintaining both versions shouldn't be a problem for now.
The only problem is that I thought "_writing a Tally program/integration for OBS Websocket 4 is very difficult, since some data isn't exposed by the API and I have to request them everytime and use lots of variables; I hope the new version is better_", but I was wrong...
Now we can make batch requests, which is great, and the permission/subscription implementation is very good and granular, but it will take some time to work on this (example: we used the `TransitionBegin` and `TransitionEnd` events to update the tally status as soon as the transition starts/ends, for long dissolvences etc, but now it returns only `transitionName`).


MatteoGheza:
Done. Now everything should work, except thing not implemented in the 5.0.0 protocol, like transition begin and transition end, source visibility, and something else.
The code, thanks to bad implementation of lots of OBSWebsocket events, has become more complex, but I tried to refactor it has much as possible and to add some code comments.

CS-5:
Since obs-websocket v5 uses port `4455` by default, it may be beneficial to add the option to reconfigure the TallyArbiter port in this PR.

Currently, this can be done with a `.env` file in the application root, but this breaks the Electron UI (as far as I can tell, the Electron UI has the port hard-coded and does not look at the environment variables - but I could be mistaken).

PR #523, opened 2022-12-21
Some UI Change on Tally View pages and chat components

chanx:
Tally View - Add Checkbox for Enable Chat/Disable Chat and automatically add ?chat=false as query params to url

Chat components can use device name instead of client only, so producer and other client can differentiate the chat is come from who.


hrueger (approved):
lgtm 👍

PR #613, opened 2023-12-21
Update angular version to 17 and do some update in UI, package

chanx:
I have update the angular to 17, solve some electron builder bug, update some package like atem-connection, electron, electron builder, fixed some other error. Please take a look and do some test for that.


chanx:
docker failed because of nodejs version, angular 17 is using node >18.13 but the installed docker is using nodejs 14.
perhaps we need to update the version the number?

PR #617, opened 2023-12-26
Fix build errors in Build Listener Clients for TTGO listeners

Jocke4f:
In the latest version (currently v2.5.34) of TTF_eSPI was a direct reference added to hal/gpio_11.h. This breaks the build of the TTGO listeners in TallyArbiter, a temporary workaround is to for the build to use v2.5.33.


Jocke4f:
I will create an issue due to this "fix" when it has been merged so that this technical debt isn't forgotten.

PR #622, opened 2024-01-08
Doc updates

Jocke4f:
Updated minimum Node.js from 14.17 to 18.13, this is due to Angular CLI dependency.
Updated OBS Studio with clarification that obs-websocket is for pre v28.


josephdadams:
@Jocke4f thoughts on including a note about OBS v28+ and the conflicting 4455 port?

Issue #603, opened 2023-10-23
[Bug]: Docker Image v3.0.4 does not start: Cannot find module bcrypt


robo-w:
### What happened?

Steps to reproduce:
* Install Docker/Podman on Rasberry Pi 4B, Raspbian 64-bit
* Run v3.0.4 image according to https://josephdadams.github.io/TallyArbiter/docs/installation/docker
* Tally arbiter does not start - see error output below

### Version

3.0.4

### Distribution

Docker

### OS

Linux

### Error stacktrace (if applicable)

```shell
Error: Cannot find module '/app/node_modules/bcrypt/lib/binding/napi-v3/bcrypt_lib.node'
Require stack:
- /app/node_modules/bcrypt/bcrypt.js
- /app/_helpers/auth.js
- /app/_helpers/config.js
- /app/index.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/app/node_modules/bcrypt/bcrypt.js:6:16)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1157:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/app/node_modules/bcrypt/bcrypt.js',
    '/app/_helpers/auth.js',
    '/app/_helpers/config.js',
    '/app/index.js'
  ]
}
```

BenJamesAndo:
I believe this has been fixed, but since no new releases have come out since the fix, then it's still not working. See discussion https://github.com/josephdadams/TallyArbiter/discussions/463

Jocke4f:
Does it work with 3.0.5?

robo-w:
> I believe this has been fixed, but since no new releases have come out since the fix, then it's still not working. See discussion #463

[...]

PR #634, opened 2024-01-17
Replacing bcrypt with bcryptjs

Jocke4f:
This change was done in:
`src/_helpers/auth.ts`
`package.json`

Replace was also made in
`Dockerfile`
but I'm not certain that we need it here. Can it be removed in this part? This can perhaps be sorted out later.

It was removed in
`.gitpod.yml`
since we should not need to rebuild this package.

The purpose is to fix https://github.com/josephdadams/TallyArbiter/issues/603.

bcrypt.js: https://github.com/dcodeIO/bcrypt.js#readme
bcrypt: https://github.com/kelektiv/node.bcrypt.js#readme


Jocke4f:
Should `package-lock.json` be updated?

josephdadams:
You can run ‘npm install’ which should update the lock file and include that with your commit

Issue #500, opened 2022-10-25
[Feature Request]: Update electron to shut down TA on MacOS shutdown/restart


josephdadams:
### What feature would you like to see added to TallyArbiter?

A user has reported that when they shut down their Mac, TA prevents the shutdown process.

We should add a listener and shut it down.

https://stackoverflow.com/questions/51804259/how-to-prevent-electron-app-from-interrupting-macos-system-shutdown

Jocke4f:
Was the fix for this included in https://github.com/josephdadams/TallyArbiter/releases/tag/v3.0.8? If so can we then close this issue?

josephdadams:
yep

PR #646, opened 2024-02-04
Allow shutdown on mac while app is running

gr8kingg:
Currently, if app is still running and the mac os is shutdown, the app prevents shutdown. This allows the shutdown to close the app.


josephdadams:
Fixes #500

Issue #640, opened 2024-01-21
[Bug]: Error [ERR_REQUIRE_ESM]


szkrstms:
### What happened?

I installed from CLI (https://josephdadams.github.io/TallyArbiter/docs/installation/cli), but but it doesn't want to start on Raspberry Pi (Linux magic-pi 6.1.21-v7+ #1642 SMP Mon Apr  3 17:20:52 BST 2023 armv7l).
~ $ sudo n 20.10.0
  installing : node-v20.10.0
       mkdir : /usr/local/n/versions/node/20.10.0
       fetch : https://nodejs.org/dist/v20.10.0/node-v20.10.0-linux-armv7l.tar.xz
     copying : node/20.10.0
   installed : v20.10.0 (with npm 10.2.3)

~ $ node --version
v20.10.0

~ $ sudo npm i -g tallyarbiter
npm WARN deprecated dgram@1.0.1: npm is holding this package for security reasons. As it's a core Node module, we will not transfer it over to other users. You may safely remove the package from your dependencies.

changed 332 packages in 5m

35 packages are looking for funding
  run `npm fund` for details

~ $ tallyarbiter 
[2024-01-21 10:29:37] info: Starting OSC Setup.
/usr/local/lib/node_modules/tallyarbiter/node_modules/esm/esm.js:1
const __global__ = this;(function (require, module, __shared__) { var __shared__;.... _(very long source code)_                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            

[/usr/local/lib/node_modules/tallyarbiter/node_modules/axios/index.js:1
Error [ERR_REQUIRE_ESM]: require() of ES Module /usr/local/lib/node_modules/tallyarbiter/node_modules/axios/index.js not supported.
Instead change the require of index.js in null to a dynamic import() which is available in all CommonJS modules.] {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.10.0

What do I wrong, or what missing?


### Version

3.0.7

### Distribution

CLI

### OS

Linux

### What browsers are you seeing the problem on?

Safari

### If applicable, What Listener Clients are You Using?

GPI

### TallyArbiter configuration

_No response_

### Relevant log output

_No response_

### Error stacktrace (if applicable)

_No response_

Jocke4f:
The issue with

> ~ $ sudo npm i -g tallyarbiter
[...]

PR #651, opened 2024-02-16
Revert axios (1.6.0 to 0.27.2)

Jocke4f:
Reverting axios (update previously made in https://github.com/josephdadams/TallyArbiter/pull/607).

In order to step to axios 1.6.0 the change in axios from Common JS to ES Module needs to be handled, this it not quickly done. This revert is only to make a release more sucessfull. A new issue will be created to track this technical debt.

When this PR has been merged to main is the suggestion that we go ahead with a new release since there are several queued up fixes that would be good to have in a formal release package.

Fixes https://github.com/josephdadams/TallyArbiter/issues/640

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
                 "@types/express": "^4.17.13",
                 "ajv": "^8.11.0",
                 "atem-connection": "3.4.0",
-                "axios": "^1.6.0",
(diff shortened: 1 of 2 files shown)


josephdadams:
There is a PR that will be submitted soon that will bring in ember+ support so let’s wait for that one before a new release

PR #652, opened 2024-02-18
Add Ember+ support as an outgoing action

netdude78:
I tested this against a Lawo console running 6.4.0.19 and one running 10.8.0.1.  

The ember path needs to be collected & verified using something like Ember+ viewer (available freely online).

Action uses a static dictionary of connections to reuse existing ember+ connections.  Connecting initially to Ember+ can result in a delay that is not great for AfV operations based on tally.


josephdadams:
@netdude78 can you update the docs too?

netdude78:
Sure. After dinner. I’ll update.

josephdadams:
Thanks!

netdude78:
I added the doc information about Ember+

PR #669, opened 2024-03-18
Added auxBux support for BMD ATEM devices

Jocke4f:
Added support for auxbux for ATEM devices. No test with real ATEM with AUX has been done.
Removed video.ME, it was replaced by video.mixEffects in change from 1.x to 2.x in atem-connection


Jocke4f:
The aux will override both preview and program. This is not an error in this code, it is a general issue that the priority handling for preview, aux and program is not implemented/working. It will therefore not be handled by this PR.

MatteoGheza (approved):

Issue #661, opened 2024-03-01
[Bug]: GPO-Listener Invert Output


DJLWegas:
### What happened?

If you invert the output in the config file and then flash a listener, the flash works but it does not end in the last state. It ends in the oposite state. If the source has Tally, the output stays LOW (should be HIGH) at the end of the flash and if the source has no Tally, the output stays HIGH (should be LOW) at the end of the flash.

### Version

3.0.8

### Distribution

Desktop

### OS

Windows

### If applicable, What Listener Clients are You Using?

GPI

Jocke4f:
Do you have the possibility to test the fix on  [Jocke4f:fix_661_gpo_invert](https://github.com/Jocke4f/TallyArbiter/tree/fix_661_gpo_invert)? At the moment don't I have anything to test with.

Jocke4f:
Fix verified with a RPI 4 with single led. This has now been solved on master when https://github.com/josephdadams/TallyArbiter/pull/675 was merged.

PR #675, opened 2024-03-23
Added in on_flash support for output_invert for lastState.

Jocke4f:
This could be a fix for https://github.com/josephdadams/TallyArbiter/issues/661. Based on code inspection is looks plausible but I don't have any board to test with at the moment.


Jocke4f:
Tested with one led on raspberry pi both "normal" and "inverted" with expected behaviour, i.e., the current state is restored after flash.

PR #685, opened 2024-04-12
Added support for m5stickC-Plus2

Jocke4f:
Improvements

- Added support for m5stickC-Plus2.
- Changed pin for tally extra preview. (G36 is input only, it was replaced by G25 but it doesn't exist on the plain m5stickC)
- Changed HIGH/LOW for external leds. (On m5stickC and m5stickC-plus is HIGH and LOW not the same for the internal led as on external leds)
- Replaced code internal defines for board types with board defines used by Arduino IDE.

Things to fix in the future:

- Fix battery level for M5stickC-Plus2.
- Add screen brightness for M5stickC-Plus2.

### DIFF
--- a/listener_clients/m5stickc-listener/README.md
+++ b/listener_clients/m5stickc-listener/README.md
@@ -39,9 +39,9 @@ Make sure you have selected the **right serial port** and the **right board type
 
 Done! Now your board is running the latest listener client firmware version. Go to the *"Setup your device"* sections to connect the board to the Tally Arbiter server.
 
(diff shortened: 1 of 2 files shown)


josephdadams:
I'm ready to merge when you are ready. I don't have an m5+2 but if it works for you, let's get it in!

Jocke4f:
I have tested with all three hw variants, i.e., m5stickc, m5stickc-plus and m5stickc-plus2.

Issue #707, opened 2024-07-24
[Bug]:  DEVICE - OSC Action - Argument not sent


AntP-vp:
### What happened?

Hi all, I'm trying to send an OSC string from a device using the Action but TA is not sending the argument part.

The monitoring in the receiver software show that no arguments 'string' is received, using the same path and argument from Companion works good and a camera view is shown in the software.

Thank you.

<img width="1280" alt="TA" src="https://github.com/user-attachments/assets/499c7a7c-def0-4607-87bf-7c0ee3c292ab">
<img width="1280" alt="Companion" src="https://github.com/user-attachments/assets/a213ed6f-0155-47d0-8652-a6860c7ced03">

### Version

3.0.10

### Distribution

Desktop

### OS

Windows

AntP-vp:
Here is the Log, all seems to be fine here.

<img width="1280" alt="TA Log" src="https://github.com/user-attachments/assets/7e889190-86b2-486c-9e79-6e7453c48dc2">

Jocke4f:
Yes, there are two bugs in OSC RunAction.

1. No data is sent.
2. String is not identified.

The PR https://github.com/josephdadams/TallyArbiter/pull/713 solves this.

PR #713, opened 2024-08-08
Fix OSC argument not sent in RunAction

Jocke4f:
Small correction in OSC RunAction:

- Added regex to identify if not float32 nor int32 then string.
- Corrected variable for oscUDP send so that data is actually sent.

This is a fix for https://github.com/josephdadams/TallyArbiter/issues/707.


josephdadams:
Thanks, @Jocke4f !

PR #717, opened 2024-08-20
Add Riedel SimplyLive as source

KSauter:
Add Riedel SimplyLive as source.

Uses TSL UDP v5 internally.

### DIFF
--- a/docs/docs/usage/sections/sources.md
+++ b/docs/docs/usage/sections/sources.md
@@ -74,3 +74,4 @@ You will need the IP address of the device, and the port (standard port is 10600
 
 ## Riedel SimplyLive
 You need to configure TSL in the SimplyLive backend to send the data to Tally Arbiter at the port you specify.
+Uses an TSL v5 UDP connection internally.


KSauter:
Draft, because i still need to test it with with the SimplyLive server. Will do that in the next 1 or 2 days and update this PR accordingly.

KSauter:
Tested with server, can confirm it works. Remove draft state, ready to merge from my side

josephdadams:
I'm ok with this even though it at first glance looks like it is just a copy of the TSL 5 UDP source type. I'm assuming you generated a unique UUID for the source type (I didn't check) :)

It would probably be a good idea to add to the docs that it is UDP in case the user does not know that and has a broader network configuration going on where UDP may not cross as intended.

KSauter:
Hi @josephdadams , thanks for your quick reply!
UUID i did generate a new one. Good hint with the UDP network, added it to the docs.


The main difference between this and plain TLS v5 UDP is the rule used to determine whether an address is visible on the pgm/prv or not.
The default TSL source uses `text tally state` (see [TSL doc](https://tslproducts.com/media/1959/tsl-umd-protocol.pdf) page 7) which does not specify anything about the tally text. The tally arbiter implementation assumes 0 => off / 1 => pgm / 2 => prv / 3 => both.
In case of simplylive, they only send 0 => off / 1 => in any pgm or prv or both.
The LH and RH tally state is correct (0 => off / non-0 => active with a color). Thats specified in the TSL doc, if a tally should be illuminated or not.

I added it to the TSL source first, including three new options:
1) mode - should color or text be used?
2) RH tally mapping - if in color mode, is RH tally pgm or prv?
3) LH tally mapping - if in color mode, is LH tally pgm or prv?
(right and left might differ on different mixers, as some of them have pgm on the right and others on the left)

I thought it would be easier for the user to have a dedicated SimplyLive source. But I can add back the different options in the TSL source if you like.

josephdadams:
Can we have both? Make it easier for the simply live users and have the flexibility for the others?

KSauter:
Sure thing. I will make a new branch with the adjustments, based on this one.

KSauter:
Hi @josephdadams 

just a quick heads up here. I didn't forget about the generic TSL v5 settings, just had too much on my plate to get my production with SimplyLive up and running.
Will implement it soon.

Best,
Kevin

PR #718, opened 2024-08-20
Fix global handling of current tally data

KSauter:
During implementation of SimplyLive source (https://github.com/josephdadams/TallyArbiter/pull/717), i noticed some strange behaviour.

E.g. the tally state in the UI was not matching the state in the VMix emulator.
I traced it to some inconsistencies in the tally state update and linked bus logic.

This PR should fix any updateDevice logic and get rid of duplicated code for linked bus logic. This relies on the internal state which is computed by "UpdateDevice" to be accurate and can be used as SSOT for all other places where the current tally state of a device is used.

### DIFF
--- a/src/_models/TallyData.ts
+++ b/src/_models/TallyData.ts
@@ -3,7 +3,7 @@ export type AddressTallyData = {
 }
 
 export type SourceTallyData = {
-    [sourceId: string]: string[];
+    [deviceSourceId: string]: string[];
 }
 
 export type DeviceTallyData = {
--- a/src/index.ts
+++ b/src/index.ts
@@ -983,51 +983,32 @@ function TSLClients_UpdateAll() {
 
 function getDeviceStates(deviceId?: string): DeviceState[] {
 	let deviceStateObj = devices.filter((d) => deviceId ? d.id == deviceId : true).flatMap((d) => currentConfig.bus_options.map((b) => {
-		const deviceSources = device_sources.filter((s) => s.deviceId == d.id);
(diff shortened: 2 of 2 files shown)


josephdadams:
@Jocke4f can you help review this?

Jocke4f:
The current code on master works for linked busses with device sources but not for Test Mode (linked busses were not properly handled). That is one error in the PR that earlier fixed the linked busses support.

Has the code been tested? I tested the code but it doesn't work for linked busses with sources nor with Test Mode. Perhaps my test was flawed but it didn't work for me.

KSauter:
I will double check.
Just to be sure how linked busses should work: A device can have more than one source, and if the device is linked to a bus, all sources for this device need to be in the recpective bus right?

josephdadams:
yes

josephdadams:
@KSauter @Jocke4f I'm in the middle of a PR adding a new source type and in the process have fixed several things related to bus options and how auxes etc. are referenced. There's been an overarching bug/design problem in TA for awhile now where instead of referring to the preview and program busses by their bus_option.id, we have been referencing them by the literal name `preview` and then doing a bus_option.type evaluation. This works ok for preview/program but when you look at the aux options, they all have the same type of `aux`.

I've switched over several places in the code to use the id instead of type, and have implemented a quick fix to replace `preview` with its actual bus_option.id (For example) which should allow all source types to work.

So please wait to merge this until that PR #726 is in place because it will affect some of what you have done here. Thanks!

PR #720, opened 2024-08-25
Change from Common JS to ES Modules

Jocke4f:
To be able to upgrade to a later axios package must the code be changed from CommonJS to ES Modules.
Axios changed from 0.28.0 to 1.7.5.


josephdadams:
I think this is a good move. I’m also open to using fetch or some other library if you think we need to move away from axios long term.

PR #795, opened 2025-03-23
Fix empty Mac-Adress | M5Atom

JxxIT:
The Mac-Adress would be 00:00:00:00:00:00, because the WiFi wasn't initialized.


Jocke4f:
What is the issue that this will solve?

JxxIT:
My M5Atom failed to read it's Mac-Adress, and defaulted to 00:00:00:00:00:00. Both of my devices had the same name and conflicted because of this.

Jocke4f:
That is interesting. I tried to reproduce it by purging my m5atom but it properly fetches the mac on startup.  Can you elaborate on the scenario, e.g., are the devices brand new or has something else been running on them. Can the problem be reproduced. Was the used sketched configured in a specific way?

JxxIT:
Hm strange, they are brand new, and I used the latest sketch. I can't test right now, I don't have them with me.

PR #800, opened 2025-04-04
Update VMix.ts

Christoph111:
reacts to ACTS commands, e.g. "ACTS Overlay1" so vMix Listener Devices work correctly (e.g. Hollyland Wireless Tally System)

response 0 ist static, because the tally status is already handled by tallyArbiter so no need to check for overlay status again.

be aware if you´re running vMix and TallyArbiter on the same machine, there will be a conflict where hollyland will connect to, so make sure to run tallyarbiter on a different computer. But since Hollyland directly works with vMix and does this great, if you only use vMix there is no need for tallyArbiter for hollyland tally system.


MatteoGheza (approved):
LGTM

PR #886, opened 2025-12-29
feat: Add MQTT integration

JonBons:
Add MQTT support to publish TallyArbiter device and bus states to MQTT brokers for integration with Home Assistant and other various platforms.

- Publish device states as JSON with full source information
- Publish individual bus states (Preview/Program/Aux) for each device
- Simple ON/OFF state topics for Home Assistant binary sensors
- Configurable broker, credentials, topic prefix, QoS, and retention
- Automatic reconnection and error handling
- Service and device availability status topics

Configuration added to config.json:
- enabled (default: false), broker (default: "localhost"), port (default: 1883), username (optional), password (optional)
- topicPrefix (default: "tallyarbiter"), retain (default: true), qos (default: 0), reconnectPeriod (default: 5000), connectTimeout (default: 10000), keepalive (default: 60), clientId (default: '', auto-generates an id of `tallyarbiter_${Date.now()}` if not set)

An example of how this could be integrated into HomeAssistant, allows you to use the data coming from TallyArbiter to link various states up to a massive number of devices (Hue lights, Zigbee, Z-Wave lights, ESPHome, etc) although not without additional delay when states change compared to directly listening to the TallyArbiter server: 
firefox_AnHeE79qGz

Another example of this being useful would be to directly integrate the MQTT data into any ESPHome compatible devices as customized listeners without require people to build custom arduino code.

Example of configuring an M5 AtomS3 Lite as an MQTT listener against messages sent to the broker from TallyArbiter: https://gist.github.com/JonBons/fc74282a4147c14d8e07c043eff061f6

### DIFF
--- a/src/_models/Config.ts
+++ b/src/_models/Config.ts
@@ -7,6 +7,7 @@ import { Device } from './Device'
 import { DeviceAction } from './DeviceAction'
 import { DeviceSource } from './DeviceSource'
 import { Source } from './Source'
+import { ConfigMQTT } from '../_modules/MQTT'
 
 export interface Config {
 	externalAddress: string
@@ -23,4 +24,5 @@ export interface Config {
 	cloud_keys: string[]
 	bus_options: BusOption[]
 	remoteErrorReporting: boolean
+	mqtt?: ConfigMQTT
 }


josephdadams:
Can you update the documentation also?

PR #1004, opened 2026-06-20
Add initial Grass Valley ETP support and updates

josephdadams:
Adds initial support for Grass Valley ETP. This was done for a client and while it worked fine for me in an emulator, they were experiencing some inconsistencies. It needs review from someone with access to a K-Frame before we can merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,6 +116,13 @@ ui-dist/*
 UI/src/app/_models
 !UI/src/app/_models/*.md
 
+# UI types (synced from src/_types, same as UI models above)
+UI/src/app/_types
+
+# UI config schema (synced from src/_helpers/configSchema.ts, same as UI models above -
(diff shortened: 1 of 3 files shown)


jreynolds67:
I have a K-Frame and can test. Are you able to give me an executable?

josephdadams:
No, but you can pull down this branch and run it from source which might be better anyway for debugging. My client was saying:

* "when I put key 1 online and then put key 2 online, and keep key 1 online, I don’t get a tally on both."
* "I have camera 1 online which has a tally, if put key 1 online it takes the tally off camera 1."

I was not able to recreate this.

jreynolds67:
Hmm okay, I only have access to a Windows machine for that K-Frame, and I have no idea how to build something from source on Windows. If you're able to give me an executable version I can test, otherwise I won't be able to help.

PR #1080, opened 2026-07-19
fix: bump trivy-action to a resolvable version

josephdadams:
## Summary

The **Analyze Docker Image** CI job (in `.github/workflows/codeql-analysis.yml`) pinned `aquasecurity/trivy-action@0.18.0`, a tag that no longer resolves. The job failed at setup on every PR with:

```
Unable to resolve action `aquasecurity/trivy-action@0.18.0`, unable to find version `0.18.0`
```

This surfaced red checks on unrelated PRs even though it's a workflow/infra issue, not a code problem.

## Change

Bump the pin to `0.36.0` (current release). All existing inputs (`image-ref`, `ignore-unfixed`, `format`, `template`, `output`, `severity`, `vuln-type`) remain valid in this version, so behaviour is unchanged apart from the action now resolving.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -69,7 +69,7 @@ jobs:
           docker build -t tallyarbiter-scan:${{ github.sha }} .
 
       - name: Run Trivy vulnerability scanner
-        uses: aquasecurity/trivy-action@0.18.0
+        uses: aquasecurity/trivy-action@v0.36.0
         with:
           image-ref: 'tallyarbiter-scan:${{ github.sha }}'
           ignore-unfixed: true


josephdadams:
The trivy version bump is confirmed working: the **Analyze Docker Image** job now resolves the action and runs all the way to the build step (previously it died in ~3s at setup).

It still shows red, but for a **different, pre-existing reason** unrelated to this PR — the UI's `npm ci` fails on an Angular peer-dependency conflict:

```
npm error While resolving: @angular/animations@21.2.9
npm error Found: @angular/core@21.2.17
npm error peer @angular/core@"21.2.9" from @angular/animations@21.2.9
```

That mismatch is present on `master` regardless of this change and should be fixed separately (align `@angular/animations` with `@angular/core@21.2.17`, or refresh the lockfile). This PR still removes the original action-resolution failure that was breaking the job on every PR.

PR #1085, opened 2026-07-23
M5StickS3 version of the TallyArbiter listener

robsealey:
I needed a version of the listener for the M3StickS3 so with (a lot of) help from ChatGPT and some testing and debugging by me, here's the working result. I hope it's of use to others.


josephdadams:
Thanks for this @robsealey — it's merged and will ship in v3.3.0.

It landed as e62b61e rather than through this PR, because the sketch needed renaming and I wanted the supporting files to go in alongside it. Your code is byte-for-byte as you submitted it; the only change is the filename, `TallyArbiter_M5StickS3_v8.ino` → `M5StickS3.ino`. arduino-cli resolves a sketch as `<folder>/<folder>.ino`, so under the original name neither the Arduino IDE nor CI could open it.

Added on top of your sketch:

- A `README.md` matching the other listener clients.
- A section in the [listener clients docs](https://github.com/josephdadams/TallyArbiter/blob/master/docs/docs/usage/sections/listener-clients.md).
- A CI job, so M5StickS3 firmware is built and attached to releases like the other boards. It builds with `PartitionScheme=default_8MB` — the board has 8 MB of flash but defaults to the 4 MB layout, which leaves only 1.25 MB for the app. `default_8MB` gives 3.19 MB and keeps the second OTA partition, which `huge_app` would have dropped.

Because of the rename this doesn't show against your name in the commit history, which I'm not thrilled about — you're credited by name in the v3.3.0 release notes instead.

Two things worth a look if you'd like to follow up, neither of which held up the merge:

1. The sketch registers with `supportsChat: true` but never handles the `messaging` event. The server takes that flag at face value and joins the listener to the messaging room, so producer messages will be delivered and silently dropped. Either add a handler or set the flag to `false`.
2. It polls `device_states` every 250 ms. The other listeners are purely push-driven, so I'd be curious what you saw that made the fallback necessary — if pushes weren't arriving, that's a server-side bug worth fixing at the source rather than polling around.

Thanks again for the contribution.

PR #1160, opened 2026-07-29
fix: stop startup crashes from the logger TDZ and unusable sources

josephdadams:
## Problem

`npm start` (`ts-node-dev src/index.ts --dev`) crashed at import time:

```
ReferenceError: Cannot access 'io' before initialization
    at logger (src/index.ts:1586)
    at Object.<anonymous> (src/index.ts:95)
```

`logger()` guarded its socket emit with `typeof io !== 'undefined'`, but `io` was a module-level `const` declared ~40 lines *below* the first `logger()` call. `typeof` does **not** protect against the temporal dead zone of a `const`/`let` binding — it throws `ReferenceError` rather than returning `'undefined'`. The guard was therefore ineffective, and the dev-mode banner at line 95 killed the process on every dev boot.

## Changes

**Logger TDZ crash**
- Declare `let io: socketio.Server` above the first possible `logger()` call, and assign it where the http server is built. The binding is now initialized-to-`undefined` from the top of the module, so `if (io)` is a real guard protecting *every* early `logger()` call, not just the dev-mode one.

**One bad source no longer takes down the server**
- `loadConfig()` logs and skips a source that fails to initialize instead of aborting startup. A config can legitimately outlive the source type it references (a type removed from this build, or one that lives on another branch); the remaining sources are still usable.
- `TallyArbiter_Add_Source()` validates the source type *before* adding it. That path runs inside `TallyArbiter_Manage()`'s promise chain, and the socket handler attaches no `.catch` — so a throw became a fatal unhandled rejection on Node 15+ and left the caller hanging. Checking first also keeps an uninitializable source out of the saved config.

`initializeSource()` itself still throws: the test-mode caller immediately dereferences its return value, and `TESTMODE` is a built-in type whose absence is a genuine programming error.

## Verification

- Original repro no longer throws; the server boots through `startUp()`.
- Against a config containing a source type not present in this build, startup logs `Source: <name> could not be initialized and will be skipped.` and continues to `Tally Arbiter running on port 4455`, process alive.
- Clean-config boot verified against a throwaway app data folder.
- `npx tsc --noEmit` — exit 0.
- `npm run format:check` — clean.

## Note

A skipped source remains in `sources` and will appear in the UI as configured-but-not-connected. That's deliberate — dropping it from the config would discard user data — but flagging it as errored in the UI would be a reasonable follow-up.


josephdadams:
@copilot resolve this conflict

PR #1161, opened 2026-07-29
fix: resolve all 17 open Dependabot alerts

josephdadams:
Closes all 17 open Dependabot alerts — 2 in the root lockfile, 15 in `UI/package-lock.json`.

Approach: non-force `npm audit fix` in the UI, plus narrowly scoped `overrides` for the copies whose parents pin them too tightly for npm to move on its own.

## Server — `package.json`, `package-lock.json`

Both root alerts were the *same* `ws` copy. `osc` pins `ws` to an exact `"8.18.0"`, so `npm audit fix` could never touch it. A scoped override lifts just that copy:

| package | before | after | advisories |
|---|---|---|---|
| `osc > ws` | 8.18.0 | 8.21.1 | GHSA-96hv-2xvq-fx4p, GHSA-58qx-3vcg-4xpx |

The override is scoped to `osc` rather than global on purpose: the remaining top-level `ws@7.5.13` serves `obs-websocket-js ^7.2.0`, sits outside both advisory ranges (`>= 8.0.0`), and a global override would force it across a major for no security gain.

## UI — `UI/package.json`, `UI/package-lock.json`

`npm audit fix` resolved these transitively, with `package.json` untouched:

```
fast-uri        -> 3.1.4     ws          -> 8.21.1
undici          -> 6.28.0    ip-address  -> 10.3.1
brace-expansion -> 1.1.16 / 5.0.8
```

The remainder were pinned by the Angular toolchain and needed overrides:

| package | before | after |
|---|---|---|
| `postcss` | 8.5.12 | 8.5.24 |
| `webpack-dev-server` | 5.2.5 | 5.2.6 |
| `@babel/core` | 7.29.0 | 7.29.7 |
| `uuid` | 8.3.2 | 11.1.1 |
| `@hono/node-server` | 1.19.17 | 2.0.12 |

`uuid` and `@hono/node-server` cross a major, so both were checked against their actual consumers rather than assumed safe:

- **uuid** — `sockjs` calls `require('uuid').v4` with no `buf` argument, so it hits neither the advisory's v3/v5/v6 `buf` path nor the v9 default-export removal. Verified it loads and generates ids under uuid 11.
- **@hono/node-server** — `@modelcontextprotocol/sdk`'s `serve` import is still exported by 2.x, and `ng version` runs clean.

Both are build-time tooling and absent from the shipped bundle: the built `main` chunk hashes identically before and after these two overrides.

## Remaining findings, deliberately left

`npm audit` still reports 17 root / 10 UI findings. **Every one traces to a single advisory**, [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg) (brace-expansion DoS) — newer than, and distinct from, the two brace-expansion alerts fixed above, and not among the 17.

Its only patched line is 5.0.8+, and 5.x exports `{ expand }` instead of being directly callable, so forcing it breaks `minimatch@3`'s `expand(...)` call. It enters the tree solely through devDependencies — `electron-builder` and `ts-node-dev` on the server, `karma` on the UI — so it is packaging and test tooling, not anything reachable from a tally operator's network. Clearing it means replacing karma and electron-builder, which is its own change rather than a lockfile bump.

## Verification

- Clean `npm ci` in both packages — exit 0
- `npx tsc --noEmit` — exit 0
- `npm run format:check` — clean
- UI production build — succeeds, bundle unchanged
- `ng serve` — boots and serves HTTP 200, no sockjs/ws errors

## ⚠️ Before release

`osc` is in the **live tally path** and its `ws` moved 8.18.0 → 8.21.1. This has not been smoke-tested against real OSC hardware — worth a run against live gear first. Same caveat the previous lockfile commit carried.


josephdadams:
@copilot resolve this conflict.

PR #1200, opened 2026-08-16
fix(relay-listener): clean install fails, reconnect goes inert, and mDNS discovery is broken

goseid:
## What does this change do?

Fixes four defects in `listener_clients/relay-listener`: it no longer fails to install standalone, it re-registers with the server after a reconnect instead of going silently inert, and mDNS discovery works again. No server, UI, or other listener code is touched.

## Why?

No existing issues — all four were found while commissioning a relay-driven camera tally rig against a v3.3.0 server, and are described in full below. Each commit stands alone; happy to split this up if you'd rather take them separately.

### 1. `socket.io-client` is required but not declared

`index.js:5` requires `socket.io-client`. `package.json` declared `socket.io` — the **server**. Under Socket.IO v2 the server package bundled the client, so it resolved transitively; v3 dropped that bundling, so since the bump to 4.8.3 the client has been absent from a clean tree and the listener dies at startup with `MODULE_NOT_FOUND`.

This stays hidden when the listener runs inside a checkout of this repo, because Node walks up and finds `socket.io-client` in the **root** `node_modules`. It only bites people who install the listener standalone, or after the root tree is cleaned. In my case the root tree had a stale `socket.io-client@4.5.0`, so the listener was quietly running a different client version than the server ships.

```bash
cd listener_clients/relay-listener
rm -rf node_modules && npm install && node index.js
# Error: Cannot find module 'socket.io-client'
```

The `socket.io` server package is not used by this listener at all, so it is replaced rather than supplemented. Lockfile regenerated.

### 2. `listenerclient_connect` is never re-sent after a reconnect

The emit sat inline at the end of `connectToServer()`, so it ran exactly once. On the server side registration is bound to the socket — `socket.join('device-' + deviceId)`, `DeactivateListenerClient(socketId)` on disconnect, `UpdateListenerClients()` filtering on `!l.inactive` — so a reconnect brings a new socket id, leaving the client in no rooms with its records inactive. Nothing re-establishes either.

The listener stays up, logs `Connected to Tally Arbiter server.`, and never actuates a relay again. Relays freeze in whatever state they held when the link dropped.

This also brings the relay listener in line with the other three, which already register inside their `connect` handler:

| Client | connect handler | emit site |
|---|---|---|
| `gpo-listener.py` | `def connect():` line 177 | line 184 |
| `blink1-listener.py` | `def connect():` line 290 | line 293 |
| `pimoroni-blinkt-listener.py` | `def connect():` line 148 | line 151 |
| `relay-listener/index.js` | logged only | end of `connectToServer()` |

Worth flagging: this failure is silent. The process is alive, the log looks healthy, and the tally lights are simply wrong — stuck on for a camera that is off air, or stuck off for one that is live. In a live production that is worse than a crash, because nothing signals anything is wrong.

### 3. An unreadable mDNS TXT record kills the process

`service.txt.version` was dereferenced unguarded at `index.js:265`, so an advertisement that doesn't decode to the expected shape throws before any connection is attempted:

```
TypeError: Cannot read properties of undefined (reading 'startsWith')
    at index.js:265:28
    at Browser.<anonymous> (node_modules/bonjour/index.js:31:13)
```

Commit 4 fixes the decode that makes this reachable, but the guard is worth having on its own: as it stood, one unexpected packet from anything advertising `_tally-arbiter._tcp` on the network was enough to kill a tally listener outright — before any connection, so nothing in the server log would suggest a client had tried. A missing or unreadable version now logs and continues; a genuine 2.x server is still refused.

### 4. `bonjour` is unmaintained and mis-decodes the server's TXT record

This is what made #3 reachable. mDNS discovery has been broken in this listener since `bonjour` moved `^3.5.0` → `3.5.1`: `service.txt` arrives as `{"ersion":"3.3.0uuid=67faf61b"}` instead of `{version, uuid}`.

**The advertisement is not at fault.** I captured the response off port 5353 and walked the TXT rdata by hand, with no library in between:

```
TXT tallyarbiter-67faf61b._tally-arbiter._tcp.local  rdlen=28
0000  0d 76 65 72 73 69 6f 6e 3d 33 2e 33 2e 30 0d 75  |.version=3.3.0.u|
0010  75 69 64 3d 36 37 66 61 66 36 31 62              |uid=67faf61b|

string 0: prefix=0x0d (13)  "version=3.3.0"
string 1: prefix=0x0d (13)  "uuid=67faf61b"
walk consumed exactly 28/28 bytes — self-consistent
```

Two RFC 6763 strings with correct length prefixes, consuming exactly the 28 bytes the record declares. `bonjour-service` on the publishing side is doing the right thing.

The fault is in `bonjour`'s decode path, and it reproduces deterministically from those exact bytes:

- `dns-packet@5` returns TXT as an **array of chunks with the length prefixes already stripped**
- `bonjour/lib/browser.js:181-188` concatenates those chunks back together **without restoring the prefixes**, then hands the result to `dns-txt`, which expects wire format
- `dns-txt` reads the leading `'v'` (`0x76` = **118**) as a length prefix, overruns the 25 bytes that remain, and swallows the rest of the record

Feeding that same `dns-txt` the raw prefixed rdata returns `{"version":"3.3.0","uuid":"67faf61b"}` correctly, which isolates it to that concatenation.

`bonjour` has not been maintained for years. `bonjour-service` is its maintained successor and is **already a dependency of this repo** — it is what the server publishes with — so this puts both sides of the advertisement on the same library. Browsing the same live advertisement with it returns `{"version":"3.3.0","uuid":"67faf61b"}`.

Two behavioural differences had to be handled, so it isn't a drop-in swap:

- `findOne` takes `(opts, timeout, callback)`, not `(opts, callback)`. Passing the callback second silently sets the timeout to `NaN` and it never fires.
- `findOne` gives up after that timeout and calls back with `null`, where `bonjour` waited indefinitely. Giving up would be a regression — the listener and the server are commonly started together at boot, so the server may not be advertising yet. It now logs and browses again, preserving the old wait-until-found behaviour.

## How did you test it?

Platform: **Source** checkout on Windows 11, Node 24.13.1. Server: **v3.3.0 desktop build**. Listener client: **relay-listener**, driving a dcttech **USBRelay4**. Source type: ATEM.

**Reconnect (commit 2)** — listener started, confirmed driving relays, server killed and relaunched:

```
11:56:05  Disconnected from Tally Arbiter server.
11:56:38  Connected to Tally Arbiter server.
11:56:38  Relay 1 off / 2 off / 3 off / 4 off   [full state resync]
```

Before the fix, the same test produced the reconnect line and then nothing at all — no `bus_options`, no `device_states`, no relay commands, ever.

**No double registration (commit 2)** — I checked the server's own registry either side of the restart by emitting `listener_clients` and counting what came back:

```
entries for this relay listener : 4  (active: 4, inactive: 0)
active registrations per socket id: {"S_BuzEekFSzkwQUoAAAH": 4}
```

Exactly one registration per relay group on a single socket, so the fix does not double-register on first connect — Socket.IO buffers the emit made while disconnected and flushes it once. After the server restart the same four came back on a new socket id (`xcn_qo899jcHEGWuAAAD`), which is precisely what was missing before.

**Standalone install (commit 1)** — clean `npm install` in `listener_clients/relay-listener` with no root `node_modules` present, then `require('socket.io-client')` resolves. That is the exact repro above, now passing.

**mDNS end to end (commits 3 and 4)** — config switched to `useMDNS: true`, listener started: `Found TallyArbiter server using MDNS: engAtem:4455`, connected, all four relays synced. Verified against `bonjour-service@1.4.4`, which is what a fresh install of the declared range resolves to. Production config restored afterwards and byte-compared against the saved copy.

**After all of the above**, tally was punched on all four cameras on the live rig and followed correctly.

- [ ] `npm run build` succeeds — **N/A.** `build` is `tsc` over `src/`; this PR touches only `listener_clients/relay-listener/`, which that build does not compile, and the diff contains no TypeScript. Happy to run it if you'd like it on the record anyway.
- [x] I ran the change against a live Tally Arbiter instance
- [x] I verified the UI still loads and existing devices/sources still work <<<SEE NOTE — check or annotate>>>

## Checklist

- [x] The description above matches what the diff actually does
- [x] The diff contains only changes relevant to this PR — no unrelated reformatting, style rewrites, or drive-by edits. `index.js` has pre-existing Prettier drift on `main`; I deliberately left it alone rather than reformatting the file, so the diff stays on the lines that changed.
- [x] Any claim I make in this description (tests pass, scan is clean, build is green) is something I actually ran and can point to

### DIFF
--- a/listener_clients/relay-listener/index.js
+++ b/listener_clients/relay-listener/index.js
@@ -8,7 +8,9 @@ const path = require('path')
 const { v4: uuidv4 } = require('uuid')
 const config_file = './config_relays.json' //local storage JSON file
 const USBRelay = require('@josephdadams/usbrelay')
-const bonjour = require('bonjour')()
+const { Bonjour } = require('bonjour-service')
+const bonjour = new Bonjour()
+const mdns_discovery_timeout = 10000 //how long each MDNS browse waits before starting over
 
 //var relay = 		null;
 var detectedRelays = []
@@ -263,7 +265,23 @@ function openSocket() {
 	}
 
 	if (server_config.useMDNS) {
-		bonjour.findOne({ type: 'tally-arbiter' }, function (service) {
+		findServerUsingMDNS()
+	} else if (ip) {
+		logger(`Reading TallyArbiter server connection info from config: ${ip}:${port}`, 'info')
+		connectToServer(ip, port)
+	}
+
+	function findServerUsingMDNS() {
+		//findOne gives up after the timeout and calls back with null, where the old bonjour
+		//package waited indefinitely. Keep browsing rather than giving up: the listener and
+		//the server are often started together, so the server may not be advertising yet.
+		bonjour.findOne({ type: 'tally-arbiter' }, mdns_discovery_timeout, function (service) {
+			if (!service) {
+				logger(`No Tally Arbiter server found using MDNS yet. Still looking.`, 'info')
+				findServerUsingMDNS()
+				return
+			}
+
 			ip = service.host
 			port = service.port
 			const advertisedVersion = service.txt ? service.txt.version : undefined
@@ -279,9 +297,6 @@ function openSocket() {
 			logger(`Found TallyArbiter server using MDNS: ${ip}:${port}`, 'info')
 			connectToServer(ip, port)
 		})
-	} else if (ip) {
-		logger(`Reading TallyArbiter server connection info from config: ${ip}:${port}`, 'info')
-		connectToServer(ip, port)
 	}
 }
 
--- a/listener_clients/relay-listener/index.js
+++ b/listener_clients/relay-listener/index.js
@@ -193,6 +193,7 @@ function connectToServer(ip, port) {
 
 	socket.on('connect', function () {
 		logger('Connected to Tally Arbiter server.', 'info')
+		registerListenerClients()
 	})
 
 	socket.on('disconnect', function () {
@@ -234,7 +235,10 @@ function connectToServer(ip, port) {
 			}
 		}
 	})
+}
 
+function registerListenerClients() {
+	//(re)register every relay group with the server; runs on each connect, including reconnects
 	for (let i = 0; i < relay_groups.length; i++) {
 		socket.emit('listenerclient_connect', {
 			deviceId: relay_groups[i].deviceId,
--- a/listener_clients/relay-listener/index.js
+++ b/listener_clients/relay-listener/index.js
@@ -266,9 +266,12 @@ function openSocket() {
 		bonjour.findOne({ type: 'tally-arbiter' }, function (service) {
 			ip = service.host
 			port = service.port
-			if (service.txt.version.startsWith('2.')) {
+			const advertisedVersion = service.txt ? service.txt.version : undefined
+			if (typeof advertisedVersion !== 'string') {
+				logger(`Could not read a version from the server advertisement, continuing anyway.`, 'error')
+			} else if (advertisedVersion.startsWith('2.')) {
 				logger(
-					`Error connecting to Tally Arbiter: Tally Arbiter server version ${service.txt.version} is not supported.`,
+					`Error connecting to Tally Arbiter: Tally Arbiter server version ${advertisedVersion} is not supported.`,
 					'error',
 				)
 				process.exit(3)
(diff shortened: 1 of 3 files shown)


goseid:
The failing `welcome` check is unrelated to this PR — `actions/first-interaction@v3` renamed its inputs to underscores and the workflow on `main` still uses the hyphenated spellings, so it fails on any first-interaction event regardless of the diff. Looks like #1173 already fixes it.

PR #1208, opened 2026-08-27
fix(relay-listener): depend on socket.io-client, handle SIGTERM, document systemd

josephdadams:
## Why

Three separate problems, all hit by anyone trying to run the relay listener as a service on a Pi.

**1. The package has never been installable on its own.** `index.js` does `require('socket.io-client')`, but `package.json` declared the *server*, `socket.io@4.8.3`. socket.io v4 does not depend on the client (`accepts`, `base64id`, `cors`, `debug`, `engine.io`, `socket.io-adapter`, `socket.io-parser` only), and `socket.io-client` appears nowhere in `package-lock.json`. A clean `npm install` followed by `node index.js` fails with `Cannot find module 'socket.io-client'`. This has been the case since the initial commit — existing installs presumably had the client hoisted in from elsewhere.

**2. Nothing cleaned up on exit.** Only `SIGINT` was handled, so `systemctl stop` / `systemctl restart` (which send `SIGTERM`) skipped the bonjour teardown entirely. And no path — signal or otherwise — ever turned the relays off, so a tally light could stay lit after the service went away, until someone power-cycled the board.

**3. The README recommended pm2.** For a single always-on Node process on a Pi, that's a second supervisor on top of the one the OS already has, and it gives no way to drop privileges — so the listener ends up running as root just to reach a USB HID device.

## What changed

- **`package.json`** — `socket.io` → `socket.io-client`, same `4.8.3` version the server side uses. Lockfile regenerated.
- **`index.js`** — added `turnOffAllRelays()` and a re-entry-guarded `shutdown(signal)`, wired to both `SIGINT` and `SIGTERM`. It turns off the relays this listener manages, destroys the bonjour instance, then exits 0.
- **`README.md`** — replaced the pm2 section with systemd: a `udev` rule for the relay (covering both the `usb` and `hidraw` subsystems, since `node-hid` can be backed by either) so the service runs as an unprivileged `tally` user, a unit file, and the surrounding setup.

Two things the README now calls out explicitly, because both are easy to get wrong:

- `WorkingDirectory=` is **required**, not cosmetic. `config_file` is the relative `'./config_relays.json'`, and `saveConfig()` writes a generated `clientUUID` back to it on first run — so the folder must also be writable by the service user. That's why the unit uses `ProtectSystem=full` rather than `strict`.
- Node.js 24 publishes no official 32-bit ARM (`armv7l`) builds, which caps a 32-bit install at Node 22 and rules out the armv6 Pi Zero / Zero W entirely. The "Relay Hardware" section's Pi Zero reference is updated to the Zero 2 W.

## Deliberately left alone

The `disconnect` handler still only logs — it does **not** turn the relays off. Dropping every tally light on a brief network blip seemed worse than holding the last known state, but that's a judgment call worth a second opinion.

Also worth considering separately: `node-hid@^2.1.1` resolves to 2.2.0, which fetches its prebuilds from GitHub at install time, so a firewalled Pi silently falls back to compiling from source. `node-hid@3.4.0` bundles prebuilds inside the npm tarball (`prebuilds/HID-linux-arm64/node-napi-v4.node`), needing neither a network fetch nor a toolchain. It would need an `overrides` entry since `@josephdadams/usbrelay` pins `^2.1.1`, plus real hardware testing — out of scope here.

## Testing

- `npm install` from clean, then all five top-level requires resolve.
- Listener starts, loads config, and stays up (falls back to the relay simulator on a machine with no board attached).
- `SIGTERM` now runs the shutdown path — logs `Received SIGTERM. Shutting down.` and exits 0. Before this change it died on the default handler.
- `npm run format:check` passes.

**Not tested on real hardware.** The relay-off path was exercised only against `USBRelaysimulator`, so the `setState` calls to a physical board on shutdown still need a check on an actual Pi with a relay attached.

Note: this repo's `.eslintrc.js` is the legacy format and the pinned ESLint is v9, so `npx eslint` won't run on these files. Pre-existing, untouched here — and the relay-listener analysis job in `build-listener-clients.yaml` is commented out anyway.

### DIFF
--- a/listener_clients/relay-listener/README.md
+++ b/listener_clients/relay-listener/README.md
@@ -23,7 +23,7 @@ The software is written in Node.js and is therefore cross-platform and can be ru
 **RUNNING DIRECTLY WITHIN NODE:**
 
 1. Install Node.js if not already installed. <https://nodejs.org/en/download/> If installing on a Windows PC, ensure that the option to install Tools for Native Modules is selected otherwise the installation of Tally Arbiter Relay Listener will fail.
-1. If installing on a Pi, run `sudo apt install libudev-dev libusb-1.0-0-dev`: The `libusb` library is necessary to communicate with the USB relay.
+1. If installing on a Pi, run `sudo apt install libusb-1.0-0`: this shared library is necessary to communicate with the USB relay. `node-hid` ships prebuilt binaries for 64-bit ARM, so no compiler is needed.
 1. Download the Tally Arbiter source code.
 1. Open a terminal window and change directory to the folder where you placed the source code.
 1. Type `npm install` to install all necessary libraries.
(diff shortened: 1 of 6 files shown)


josephdadams:
@copilot resolve the merge conflicts in this pull request

Issue #1205, opened 2026-08-20
[Bug]: webpage appear freeze


tetotaglia:
### What happened?

The producer webpage appear freeze, don't update with the change on mixer video.
With a refresh sometime appear "No devices are available for tally monitoring at this time."
Web client don't update the status, need to refresh manually.

### Version

3.3.0

### Distribution

Docker

### OS

Linux

### What browsers are you seeing the problem on?

Chrome

### If applicable, What Listener Clients are You Using?

Web (Chrome)

### TallyArbiter configuration

```javascript
{
  "security": {
    "jwt_private_key": "aQo1uFYYr+c53wL5iH7RMivrHx7PU4ozBXLdstM6szc0ahB82cchMkwpW1XMDy9H4fUxTP6qVrerux0kE4OcGV+/m7o6yhvGLHTKKOAtiFeZFvJNRQieeW1SrYm/JMKNJ1QiPDFOYPG4h/0GbOfcfUllOQHXdNwoyF0OzV47s6OBCcXeCcnW/H2hwc/BXL31iO0riQ7RJbQRaCWwPP5iQxj8aqqMDShiMYp4GmDfXXnzan8IgHW5gp5Aw94moTrPNYpP2YAInZN1zNodOodM2F+Sg1pBUfIyWAUeE2+orbAJ+cBsSchmcqpOXrJDoPoACZ7klWNfdmP0652GCZWOLw=="
  },
  "users": [
    {
      "username": "producer",
      "password": "$2a$10$3TfqLifQEimOG5rvPd7HMeOUVld0GUpXmONWbtRz6.deu6Zs9YGqS",
      "roles": "producer",
      "mustChangePassword": true
    },
    {
      "username": "admin",
      "password": "$2a$10$fLhHaVUa8Haa5EgjyyfQiOW1H5H5yt9PX4WoGdNKwY2oFzEVRNPaq",
      "roles": "admin"
    }
  ],
  "cloud_destinations": [],
  "cloud_keys": [],
  "device_actions": [],
  "device_sources": [
    {
      "deviceId": "2984c589-0fb9-4709-9158-9ecbaf414d25",
      "sourceIdx": "0",
      "address": "1",
      "sourceId": "9f6b6475-56e2-4a6d-9c76-5523ec9cd2ad",
      "id": "53e75bea-79fd-4122-bf1d-f073da36e710"
    },
    {
      "deviceId": "53a7bdff-73b6-4663-a3dd-492338500caf",
[...]

PR #1212, opened 2026-08-29
feat(sources): add Pixelhue Q8 source

goseid:
## What does this change do?

Adds a **Pixelhue Q8** source type. The processor has no tally or UMD protocol, so tally is derived from its layer model: an input is on program when an enabled layer carries it on the program scene of a screen being followed, and on preview when the same is true of the preview scene.

New file `src/sources/PixelhueQ8.ts` plus a section in the sources documentation. Nothing existing is modified.

## Why?

No existing issue — I needed it for a rig where a Q8 is the final switcher in front of an ATEM, and camera tally has to follow what actually reaches the audience screens.

A few things about the device are worth knowing, because they shaped the implementation and none of them are guessable from the outside:

**It has no tally protocol.** Screens hold a program scene and a preview scene, each a set of layers, and every layer names the input it displays. That is the only source of truth, so the source reads `screen/list-detail` and `layers/list-detail` and derives tally from them.

**Only some screens should drive tally.** A Q8 exposes real screens, auxes and a multiviewer. The multiviewer carries *every* input permanently, so following it would put the whole rack on program. It is excluded structurally via `screenIdObj.type` rather than by name. Which of the remaining screens matter is site-specific — on my rig two of five drive camera tally — so it is a config field, and blank follows all of them.

**Authentication needs no password.** `GET /unico/v1/node/open-detail` returns the serial and the device's boot time unauthenticated, and the token is the serial signed with the boot time. Because it is keyed to boot time, a device reboot silently invalidates it — the device reports that as an error code inside an HTTP 200 body — so the source re-mints on every connect and retries once on that code.

**Both sides of a fade are on air.** An input fading in is on screen from the moment the fade starts, and one fading out stays there until it ends. The device does not report it that way: it moves program to the new scene partway through the effect. Reading once after everything settles gets the outgoing case right by accident and lights the incoming camera a whole fade late — which a camera operator experiences as being live before their tally says so. The take frame carries the screens involved, the effect duration, and a status that opens the transition and another that closes it, so the source publishes the union of outgoing and incoming program for the duration and settles when the device says the take is done.

**Addresses are the connector position.** An operator picking a tally source recognises the name they gave a connector or the port it is plugged into, not the composite id the layer records use internally. So inputs are addressed as `In 1-11` — card 1, port 11, exactly as the device's UI, its documentation and its own default connector names write it — and labelled `Cam 1 (In 1-11)`. Position beats the connector name as an address because it survives a rename, and beats the internal id because it is what the operator reads off the device. Cards and ports are 0-indexed in the API and 1-indexed everywhere a person looks, which is the trap.

**Reads are served on both edges of a burst.** A take arrives as several frames over about 250ms and the device applies the swap on the last of them, having first reported the scene it is rebuilding as briefly empty. A fixed delay anchored to the first frame either races the swap or over-waits by the length of the burst, and the latter is felt directly as tally lag — 600ms measured as "about a second" by the operator using it. So the first frame of a burst is served immediately and a confirming read follows once the device goes quiet, with a cap so that dragging a layer in the device's own UI cannot postpone tally indefinitely.

The `Screens` field declares help text. That renders once the companion UI fix lands (`fix(ui): show config field help and address labels in source settings`, submitted separately) — the source dialog does not currently display `field.help` for any source type. The same information is in the documentation either way.

Verified against a **Q8**. The vendor's own Companion module maps the P10, P20 and P80 to identical endpoints, but I have no access to those, so the source is scoped and named for the Q8 rather than claiming a family it has not been run against.

## How did you test it?

Platform: **Source** checkout on Windows 11, Node 24.13.1 (Angular 22's CLI needs >= 24.15.0, so UI builds used a portable Node 24.19.0). Hardware: a **Pixelhue Q8** (`modelId 29974`, firmware serving the API on 8088) with 48 input connectors across four cards, six cameras on 12G-SDI, and five real screens plus a multiviewer. Tally output driven through the **relay listener** to physical camera tally lamps, with a **Blackmagic ATEM** configured as a second source on the same devices throughout.

**A full production show.** The source ran for 5h15m unbroken — 1893 samples at 10s intervals — with **zero fetch errors, zero fatals and no reconnects**. Over the 2h31m main act an ATEM feed was on the tallied screens 91.3% of the time and direct camera looks 1.4%; all six cameras tallied correctly, and the two mapped to relays lit real lamps.

**Directed testing afterwards**, sampling both the device and Tally Arbiter's own `device_states` at 250ms and snapshotting the device on every notification frame:

- **Program, preview, preview-only and clear-down** all resolve correctly, and cameras resolve independently rather than in lockstep.
- **Screen filtering** — with two of five screens configured, layers on the other three produce no tally. Blank follows all real screens. The multiviewer never tallies.
- **Transitions** — a take decomposes into three frames over ~250ms; the empty-scene window during a rebuild is ~130ms and only ever affected the preview bus, never program. After the transition handling went in, the operator's verdict on fades was "they come on immediately at the start of the fade in and stay lit until the fade is complete… perfectly snappy".
- **Latency** — first frame to tally change measured at 386ms with a 250ms trailing debounce and 676ms with 600ms, which is what drove the move to a leading edge. Endpoint response times: `layers/list-detail` 16ms (192KB), `screen/list-detail` 6ms, `interface/list-detail` 35ms (498KB, so it is not re-read on every transition).
- **Addresses** — all 48 input connectors produce distinct positions, no collisions. Confirmed end to end: the device reported program on screen 6 as its internal `2:11`, which resolved to address `In 2-11`, matched the configured device source and lit the mapped device.
- **Reboot handling** — the rejected-token path was exercised by the source re-minting after the device's boot time moved.

- [x] `npm run build` succeeds — ran on this branch, exit 0. `prettier --check` passes on both files.
- [x] I ran the change against a live Tally Arbiter instance
- [x] I verified the UI still loads and existing devices/sources still work

On the last box: the ATEM source and its six device sources ran alongside this one for two days on the same instance and continued to report normally, and the Sources & Devices tab lists and edits both. Sources, devices and device sources were added and edited through the settings dialog during that period.

One unrelated observation while testing, in case it is useful: on the first load after signing in, the Settings page can sit on the loading spinner until you move to another tab and back. It reproduces on unmodified `main`, and I have put what I found on #1205 — it looks like room membership and the access token both being keyed to the socket id, with nothing re-joining after a reconnect. It has nothing to do with this PR.

## Checklist

- [x] The description above matches what the diff actually does
- [x] The diff contains only changes relevant to this PR — it adds one new source file and one documentation section, and modifies nothing else. No reformatting, no drive-by edits.
- [x] Any claim I make in this description (tests pass, scan is clean, build is green) is something I actually ran and can point to

### DIFF
--- a/docs/docs/usage/sections/sources.md
+++ b/docs/docs/usage/sections/sources.md
@@ -17,6 +17,7 @@ The following source types are supported:
 - OBS Studio
 - Open Sound Control (OSC)
 - Panasonic AV-HS410
+- Pixelhue Q8
 - Riedel SimplyLive
 - Roland Smart Tally
 - Roland VR-50HD-MKII
@@ -83,6 +84,16 @@ Community project: [GPI-to-OSC](https://github.com/simply-Gamic/GPI-to-OSC) conv
 ## Panasonic AV-HS410
 
 You will need the IP address of the switcher. Multicast must also be enabled on the switcher and your network in order to receive the tally data, therefore Tally Arbiter and the Panasonic device must reside on the same subnet.
+
+## Pixelhue Q8
+
+You will need the IP address of the processor and its API port, which is 8088. No password is required. Source addresses are the input number as the device reports it, and the address list is populated with the input names configured on the device, so inputs named `Cam 1`, `Cam 2` and so on appear under those names.
+
+The Q8 has no tally or UMD protocol, so tally is derived from its layer model. An input is on program when an enabled layer carries it on the program scene of a screen being followed, and on preview when the same is true of the preview scene. Because a screen can show several layers at once, more than one input can be on program simultaneously, which is normal for a multi-box or picture-in-picture layout.
+
+Leave **Screens** blank to follow every screen, or list the screens that should drive tally by name or number, separated by commas, for example `Portrait HL, Portrait HR`. The multiviewer is never followed, whether or not it is listed, because every input is present on it permanently and following it would put the entire rack on program.
+
+Tally reflects whether a layer is enabled rather than whether it is ultimately visible, so an input on an enabled layer that happens to sit behind another layer still reports as on program.
 
 ## Riedel SimplyLive
 
--- a/docs/docs/usage/sections/sources.md
+++ b/docs/docs/usage/sections/sources.md
@@ -93,6 +93,8 @@ The Q8 has no tally or UMD protocol, so tally is derived from its layer model. A
 
 Leave **Screens** blank to follow every screen, or list the screens that should drive tally by name or number, separated by commas, for example `Portrait HL, Portrait HR`. The multiviewer is never followed, whether or not it is listed, because every input is present on it permanently and following it would put the entire rack on program.
 
+During a fade, both the outgoing and the incoming inputs are reported on program for as long as the transition lasts. An input fading in is on screen from the moment the fade begins, so its tally lights then rather than when the fade finishes, and an input fading out keeps its tally until the fade is complete. A cut has no such window and takes effect immediately.
+
 Tally reflects whether a layer is enabled rather than whether it is ultimately visible, so an input on an enabled layer that happens to sit behind another layer still reports as on program.
 
 ## Riedel SimplyLive
--- a/docs/docs/usage/sections/sources.md
+++ b/docs/docs/usage/sections/sources.md
@@ -87,7 +87,11 @@ You will need the IP address of the switcher. Multicast must also be enabled on 
 
 ## Pixelhue Q8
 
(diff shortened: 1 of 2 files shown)


goseid:
The timer finding was a real bug and is fixed in 7c82bd6 — the fade duration off the wire is now clamped.

The certificate one I've left, with the reasoning in a comment at the call site. The device serves that socket with a self-signed certificate and it's the only notification channel it offers — plain ws on that port is refused, and the websocket it advertises on its API port doesn't serve the path. It also gives up nothing already protected: all tally data is read over plain HTTP from the same device, and the token is derived from values the device publishes unauthenticated.

goseid:
I think I've got the timer licked now.

The duration is now clamped immediately before the setTimeout, so the timer is bounded to 10s regardless of what the device reports. If the query still flags it, it's objecting to the value originating remotely rather than to it being unbounded — the alternative would be ignoring what the device says the fade length is, which would break the feature.

